PR #5, opened 2019-01-17
Add auto deploy script for docker: start app and postgres.

zongzw:


zongzw:
Agree. Left this pull request here to reimplement it with docker-compose later.

zongzw:
That's right.

PR #6, opened 2019-01-18
fix travis error of call-me-maybe dependency.

zongzw:


zhaoqin-github:
Open an issue for this bug #7 

This code change can workaround this bug. Need to look for a better way.

zhaoqin-github:
Oas-validator fixes the bug in version 2.0.2. We can revert this workaround now.

PR #8, opened 2019-01-18
Add waf/vs/policy concepts.

zongzw:
Include: model/datasource/repository/controller


app/waf/src/controllers/application.controller.ts
```diff
@@ -0,0 +1,138 @@
...
+import {Application} from '../models';
+import {ApplicationRepository} from '../repositories';
+
+export class AdcController {
+  constructor(
+    @repository(ApplicationRepository)
+    public applicationRepository: ApplicationRepository,
+  ) {}
+
+  @post('/Applications', {
```
zhaoqin-github: Suggest not to use upper case letter in url path.

app/waf/src/models/application.model.ts
```diff
@@ -0,0 +1,51 @@
+import {Entity, model, property} from '@loopback/repository';
+
+@model()
+export class Application extends Entity {
+  @property({
+    type: 'number',
```
zhaoqin-github: Id should be string type. We will need to generate string type uuids for them.

app/waf/src/models/wafpolicy.model.ts
```diff
@@ -0,0 +1,43 @@
+import {Entity, model, property} from '@loopback/repository';
+
+@model()
+export class Wafpolicy extends Entity {
+  @property({
+    type: 'number',
```
zhaoqin-github: Id should be string type. We will need to generate string type uuids for them.

app/waf/src/models/application.model.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  @property({
+    type: 'string',
+  })
+  status?: string;
+
+  @property({
+    type: 'date',
+  })
+  create_time?: string;
```
zhaoqin-github: I think 'created_at' should be a more popular name for this property.

app/waf/src/models/application.model.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  @property({
+    type: 'date',
+  })
+  update_time?: string;
```
zhaoqin-github: I think 'updated_at' should be a more popular name for this property.

app/waf/src/models/application.model.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  @property({
+    type: 'number',
+  })
+  policy_id?: number;
```
zhaoqin-github: Renaming it to wafpolicy_id? There are several types of 'policy' in BIG-IP. I feel renaming it to wafpolicy_id will avoid naming conflict, and will be consistent with you wafpolicy model.

app/waf/src/controllers/application.controller.ts
zongzw: typo. fixed.

app/waf/src/models/application.model.ts
zongzw: Agree!

app/waf/src/models/application.model.ts
zongzw: Agree.

app/waf/src/models/wafpolicy.model.ts
zongzw: Agree.

zongzw:
Deliberate review! Accepted all.

PR #9, opened 2019-01-21
Enable docker deploying method for development.

zongzw:


app/waf/deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,57 @@
+version: "3"
+services:
+  waf:
+    image: wafaas-env:latest
+    build:
+      context: .
+    container_name: wafaas-dev-container
+    command: ["bash", "-c", "cd /root/app_dir && npm install && npm start"]
```
zhaoqin-github: The docker image should ship node_moudels, right?

zongzw: yes. the 'command' will run every time when 'docker-compose'. 'node_modules' would be shipped on host's directory(in the source code folder).

zongzw:
> Some suggestion about the path name. My original idea is to create one Dockerfile under each app/xxx folder for that app, and create one script or yaml file under top folder (or deploy folder) to deploy all those apps. Do you think that make sense?

Right, that should be a more regular way.  Do we need to do it right now? Or we can make another pull request to get it?

app/waf/deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,57 @@
...
+    links:
+      - postgres:postgres-server
+      - AS3:as3-server
+
+  postgres:
+    image: postgres:11.1
+    container_name: postgres-server
+    ports:
+      - 5432:5432
+    volumes:
```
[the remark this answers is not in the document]
zongzw: Data in postgres is important, user will never store it into a container. 
Exposing the volume make the data backup be possible.

zongzw:
This is only for development. not for product release. Close it.

PR #14, opened 2019-01-22
Define AS3 info service

zhaoqin-github:
This code change only add AS3 info service. Other services will be added later. Since LB4 currently does not support to stub service object. We will defer the unit test of AS3 service code.

### DIFF
--- a/app/waf/src/controllers/ping.controller.ts
+++ b/app/waf/src/controllers/ping.controller.ts
@@ -1,5 +1,9 @@
 import {Request, RestBindings, get, ResponseObject} from '@loopback/rest';
 import {inject} from '@loopback/context';
+import {AS3Service} from '../services';
+
+const AS3_HOST: string = process.env.AS3_HOST || 'localhost';
+const AS3_PORT: number = Number(process.env.AS3_PORT) || 8443;
 
 /**
  * OpenAPI response for ping()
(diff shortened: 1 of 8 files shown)


app/waf/src/controllers/ping.controller.ts
```diff
@@ -1,5 +1,9 @@
 import {Request, RestBindings, get, ResponseObject} from '@loopback/rest';
 import {inject} from '@loopback/context';
+import {AS3Service} from '../services';
+
+const AS3_HOST: string = process.env.AS3_HOST || 'localhost';
+const AS3_PORT: number = Number(process.env.AS3_PORT) || 8443;
```
zongzw: We may have more configuration items in future, just thinking if we could separate them to a json-file datasource/repository/model/(controller if necessary for configurable from api) for all configuration?

app/waf/src/datasources/as3.datasource.json
```diff
@@ -0,0 +1,21 @@
...
+  "options": {
+    "headers": {
+      "accept": "application/json",
+      "content-type": "application/json"
+    },
+    "strictSSL": false
+  },
+  "operations": [
+    {
+      "template": {
```
zongzw: can we add a timeout property? 
ping for as3 information should be over very quickly. waiting for default 2min timeout may be too long..

zhaoqin-github: I do not find a way to specify timeout for outbound calls. Do you know how to specify it?

zongzw: Don't know yet.

PR #22, opened 2019-01-24
Define tenant association with ADC

zhaoqin-github:


app/waf/src/models/tenantassoc.model.ts
```diff
@@ -0,0 +1,31 @@
...
+    id: true,
+    required: true,
+  })
+  tenant_id: string;
+
+  @property({
+    type: 'string',
+    required: true,
+  })
+  adc_id: string;
```
zongzw: adcId?

zhaoqin-github: done

PR #23, opened 2019-01-24
add application.wafpolicyId check; add id auto-generation.

zongzw:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -36,7 +38,15 @@ export class AdcController {
     },
   })
   async create(@requestBody() adc: Partial<Adc>): Promise<Adc> {
-    return await this.adcRepository.create(adc);
+    if (!adc.id) {
```
zhaoqin-github: Should we refuse to input adc.id via API?

app/waf/src/models/application.model.ts
```diff
@@ -33,17 +33,17 @@ export class Application extends Entity {
   @property({
     type: 'date',
   })
-  created_at?: string;
+  createdAt?: string;
```
zhaoqin-github: Can we hold a while and carefully decide with naming format is better for us?

app/waf/src/controllers/adc.controller.ts
zongzw: Since we declare the api spec based on model definition which has all properties in it, we cannot stop user input it(generally user can remove it, or left it empty). 
if user inputs it, and in our logic, we change it to uuid or reject the request, it thus seems a bit strange. 
So, after thinking it over, I think allowing users to bring their own id is better, and at least no bad known impact.

app/waf/src/models/application.model.ts
zongzw: Not sure about it. 
Before knowing the best solution, it's better to keep all keys with same naming format. 
`{"wafpolicyId": "1234", "createAt": "2019"}` gives consistant feeling than  `{"wafpolicyId": "1234", "create_at": "2019"} `

app/waf/src/controllers/application.controller.ts
```diff
@@ -38,7 +42,18 @@ export class ApplicationController {
...
+      application.id = uuid();
+    }
+    if (application.wafpolicyId) {
+      await this.wafpolicyRepository.findById(application.wafpolicyId);
+    }
+
+    try {
+      return await this.applicationRepository.create(application);
+    } catch (error) {
+      throw new HttpErrors.BadRequest(error.detail);
```
zhang-shengping: I do not know this will affect HTTP error response such as 50x, 40x etc?

app/waf/src/controllers/adc.controller.ts
```diff
@@ -15,9 +15,11 @@ import {
   put,
   del,
   requestBody,
+  HttpErrors,
 } from '@loopback/rest';
 import {Adc} from '../models';
 import {AdcRepository} from '../repositories';
+import uuid = require('uuid');
```
zhang-shengping: is 'uuid' a javascript package rather than typescript package? why do we choose this package.

app/waf/src/controllers/application.controller.ts
zongzw: yes, here `error` is 500, but we want to to be 400 because of user's invalid input.

app/waf/src/controllers/adc.controller.ts
zongzw: it is generated by tslint, uuid should an internal npm package, better to use it..

app/waf/src/controllers/adc.controller.ts
zhang-shengping: I set the id as not required, and put a uuid when I before the data put into DB.

app/waf/src/controllers/application.controller.ts
zhang-shengping: it is possible not invalid input error in the future. we probably need to catch a more specific exception in the future.

zongzw: yes

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: I do not think allowing user to input id when creating resource is an usual way... But we can talk that later. Since we do not have a clear API definition yet, agree to merge code first and fix it when we refining the API.

app/waf/src/models/application.model.ts
zhaoqin-github: Let us merge the code now, and talk about the naming later when we refining the API.

PR #24, opened 2019-01-25
Add docker deploy staff.

zongzw:


app/waf/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ADD dist dist
+RUN rm -r dist/test
+ADD public public
+COPY package.json index.js ./
+RUN npm install -g nc
+RUN npm install
+
+ENTRYPOINT \
+    timeout=15; \
+    while [ $timeout -gt 0 ]; do \
```
zhaoqin-github: I think every app image should be independent and as simple as much. It may not assume any dependencies. Those dependencies should be configurable, and deployment tool will take care those dependencies.

app/waf/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# upload image with (authorized user only)
+#   docker push f5devcentral/f5-wafaas-openstack:latest
+
+FROM node:11.6.0
+
+LABEL maintainer="F5 Networks Inc. Corp."
+EXPOSE 3000/tcp
+
+WORKDIR /root/app_dir
+ADD dist dist
```
zhaoqin-github: Do you assume that compiling work should be done before building Docker image? Will it be better, if we compile the code during building Docker image?

app/waf/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# build with:
+#   docker build . -t f5devcentral/f5-wafaas-openstack:latest
+
+# upload image with (authorized user only)
+#   docker push f5devcentral/f5-wafaas-openstack:latest
+
+FROM node:11.6.0
+
+LABEL maintainer="F5 Networks Inc. Corp."
+EXPOSE 3000/tcp
```
zhaoqin-github: Is it configurable?

app/waf/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# build with:
+#   docker build . -t f5devcentral/f5-wafaas-openstack:latest
+
+# upload image with (authorized user only)
+#   docker push f5devcentral/f5-wafaas-openstack:latest
+
+FROM node:11.6.0
```
zhaoqin-github: What is the size of this image? Did you compare it with any other image like node:11-alpine?

app/waf/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+LABEL maintainer="F5 Networks Inc. Corp."
+EXPOSE 3000/tcp
+
+WORKDIR /root/app_dir
+ADD dist dist
+RUN rm -r dist/test
+ADD public public
+COPY package.json index.js ./
+RUN npm install -g nc
+RUN npm install
```
zhaoqin-github: npm install --production ?

app/waf/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+LABEL maintainer="F5 Networks Inc. Corp."
+EXPOSE 3000/tcp
+
+WORKDIR /root/app_dir
+ADD dist dist
+RUN rm -r dist/test
+ADD public public
+COPY package.json index.js ./
+RUN npm install -g nc
```
zhaoqin-github: Needn't this one, if we needn't to test db connection.

app/waf/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+FROM node:11.6.0
+
+LABEL maintainer="F5 Networks Inc. Corp."
+EXPOSE 3000/tcp
+
+WORKDIR /root/app_dir
+ADD dist dist
+RUN rm -r dist/test
+ADD public public
+COPY package.json index.js ./
```
zhaoqin-github: Maybe we can consider to use npm-shrinkwrap to reduce image size. Or, maybe we can consider it in the future.

app/waf/Dockerfile
zongzw: I suppose travis does the 2-phase jobs: first `npm build`, second: `docker build`. 
benefit for this is `npm build` can run on the host (instead of container) before generating the image.

app/waf/Dockerfile
zongzw: The port is only a convention, only visible in the running container. 
I think there's no need to make it configurable.

app/waf/Dockerfile
zongzw: 11.6.0 is big. That should be good to use 11-alpine after testing.

app/waf/Dockerfile
zongzw: agree.

app/waf/Dockerfile
zongzw: although we add `depends_on` in docker-compose.yml, docker-compose has no strategy to make sure to start waf app after postgres is ready. That means compose doesn't know if postgres is ok or not.
As mentioned before, better way is update our waf app to check db connection before start.

app/waf/Dockerfile
zongzw: ok for it.

app/waf/Dockerfile
zongzw: After a day of investigation, DataSource's EventEmitter would post error immediately after any action to db datasource, even get: `app.get<DataSource>('datasources.db')`. So still need more knowledge about the datasource usage, and try to optimize it in future. 

Error output: 

```
Connection fails: Error: connect ECONNREFUSED 127.0.0.1:5432
It will be retried for the next request.
events.js:173
      throw er; // Unhandled 'error' event
      ^

Error: connect ECONNREFUSED 127.0.0.1:5432
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1083:14)
Emitted 'error' event at:
```

PR #26, opened 2019-01-28
add member model and tests

zhang-shengping:


app/waf/src/datasources/as3.datasource.ts
```diff
@@ -1,14 +0,0 @@
-import {inject} from '@loopback/core';
-import {juggler} from '@loopback/repository';
```
zongzw: why delete it?

app/waf/src/controllers/member.controller.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  @post(prefix + '/members', {
+    responses: {
+      '200': {
+        description: 'Member model instance',
+        content: {'application/json': {schema: {'x-ts-type': Member}}},
+      },
+    },
+  })
+  async create(@requestBody() member: Member): Promise<Member> {
```
zongzw: As mentioned f2f, we need to chagne `Member` to `Partial<Member>` to allow user to input part of Member instance.

app/waf/src/controllers/member.controller.ts
```diff
@@ -0,0 +1,141 @@
...
+    member.id = uuid();
+    return await this.memberRepository.create(member);
```
zongzw: This would raise 500 internal server error, however, in many situation, it's not errors from server. but input error. 
It's ok for now. but should be prechecked in future.

app/waf/test/acceptance/member.controller.acceptance.ts
```diff
@@ -0,0 +1,148 @@
...
+      .query({filter: {where: {addressFamily: member.addressFamily}}})
+      .expect(200, [toJSON(member)]);
+  });
+
+  it('patch ' + prefix + '/members', async () => {
+    const memberObject = createMemberObjectWithoutID();
+
+    const member = await giveMemberData(wafapp);
+
+    // pzhang(NOTE): return a count
```
zongzw: remove it. 
Not sure if we need to add name(xxx): in code. 
Need to talk about it.

zhaoqin-github: If fact we often add name into code of open source project. :-)

zongzw: matural project needn't this. 
If you add it anywhere in the code. are you responsible for explain the code at any time?

Code itself is the good document. I prefer to not use it. We may not care who did it, but only care if it does right or not.

app/waf/src/controllers/member.controller.ts
```diff
@@ -0,0 +1,141 @@
...
+
+const prefix = '/adcaas/v1';
+
+export class MemberController {
+  constructor(
+    @repository(MemberRepository)
+    public memberRepository: MemberRepository,
+  ) {}
+
+  @post(prefix + '/members', {
```
zhaoqin-github: The member resource of lbass v2 API looks like "/v2/lbaas/pools/{pool_id}/members". Suggest to be consistent with it.

app/waf/test/acceptance/member.controller.acceptance.ts
zhaoqin-github: I will not vote 'Y' or 'N' for this. I just hope to say in very large open source project, that name will help you to look for the code owner. And another usage seems like "到此一游“

zhang-shengping: I just add the comments for other coders and myself, just in case of "only God knows" situation happens on a certain day.

app/waf/src/controllers/member.controller.ts
zhang-shengping: I just read the code, it probably needs to be added into the pool controller.

app/waf/test/acceptance/member.controller.acceptance.ts
zongzw: I'm not opponent to adding comments in code. In fact, sometimes, they are very necessary. 

What I'm concerning is marking them with (name @yyy.xx.zz) . 
First, code readers don't care who did it, only care about why.
Second, even you left your name, you may be unable to be responsible for explaining it. If you do, what's the timescope of the 'service period'?

We can discuss it later.

PR #29, opened 2019-02-01
Add under construction notice to README.md

zhaoqin-github:


zongzw:
Why do we need to add it?
This repository is placed under F5Networks, indicates that we are doing works on wafaas, doing the investigation on the possibility of wafaas on openstack. We are trying to attracting others to show here the example. 
This 'under construction' does mean 'skip us' or 'not ready'.
What's more, I think we may soon reach an archievement of showable thing ready. The 'under construction' would soon/finally removed.

zhaoqin-github:
@zongzw This code change is a 'reply' to the comment made by Gruber last evening.

zongzw:
> @zongzw This code change is a 'reply' to the comment made by Gruber last evening.

I think leaving it to reader themselves to justify the value of repo is better. 
Comparing to telling user this notice, making our process/logic stronger should be more efficient. 
I think if we guarantee every merge is Agileable, our document is right up to date, isn't it that our repo a good example for wafaas/adcaas based on lb4?

PR #31, opened 2019-02-03
Enable debug for developing.

zongzw:
Make it possible to debug the repo via step-by-step mode.

### DIFF
--- a/app/waf/tsconfig.json
+++ b/app/waf/tsconfig.json
@@ -1,4 +1,10 @@
 {
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "es5",
+    "outDir": "./dist",
+    "sourceMap": true
+  },
   "$schema": "http://json.schemastore.org/tsconfig",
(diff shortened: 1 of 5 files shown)


app/waf/tsconfig.json
```diff
@@ -1,4 +1,10 @@
 {
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "es5",
```
zhaoqin-github: I not quite sure about what will be impacted, if we specify 'ES5' or 'ES6' as the target language. Which one will be best for us?

zongzw: It seems not abvious to us since we have no webpage except /explorer. 
The 2 settings is related to webbrowser's supportability. See the link: https://blog.angularindepth.com/configuring-typescript-compiler-a84ed8f87e3, search `Transpiling`. 
To us, select 'ES5' is most stable. ES6 is somehow too new. ES3 is somehow too old(it's default).

PR #35, opened 2019-02-15
add endpointpolicy and rules

jikui:
Add the endpointpolicy and rules definitions


app/waf/src/models/wafpolicy.model.ts
```diff
@@ -17,25 +17,9 @@ export class Wafpolicy extends Entity {
...
-    type: 'boolean',
     required: true,
   })
-  shared: boolean;
-
-  @property({
-    type: 'array',
-    itemType: 'string',
-  })
-  tenant?: string[];
```
zongzw: how about keep tenant here.

PR #36, opened 2019-02-20
Deploy application configuration to AS3 gateway

zhaoqin-github:


app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +173,79 @@ export class ApplicationController {
...
+
+  @post(prefix + '/applications/{id}/deploy', {
+    responses: {
+      '204': {
+        description: 'Application deploy success',
+      },
+    },
+  })
+  async deployById(@param.path.string('id') id: string): Promise<Object> {
+    let application = await this.applicationRepository.findById(id);
```
zongzw: Possible exception...

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +173,79 @@ export class ApplicationController {
...
+
+    let params: {[key: string]: Object} = {
+      application: application,
+    };
+
+    let tenantAssoc = null;
+    try {
```
zongzw: `findById` is widely used, but it's not the best way. 
Why not use `find` with filter: 

```
    let tenantAssoc: TenantAssociation[];
    tenantAssoc = await this.tenantAssociationRepository.find({
      where: {
        tenantId: application.tenantId
      }
    });
    if (tenantAssoc.length === 0) throw new HttpErrors.UnprocessableEntity("xxxxx");
```

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +173,79 @@ export class ApplicationController {
...
+          throw new HttpErrors.UnprocessableEntity(
+            'No ADC associated with Application ' + application.id,
+          );
+        } else {
+          throw e;
+        }
+      }
+    }
+
+    let adc = await this.adcRepository.findById(tenantAssoc.adcId);
```
zongzw: Use `find` or you need to `try...catch` the possible exception.

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +173,79 @@ export class ApplicationController {
...
+    params.adc = adc;
+
+    if (application.services.length === 0) {
```
zongzw: This is a validation, need to move to the beginning of the function body once after 'found that application'.

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +173,79 @@ export class ApplicationController {
...
+      params.service = service;
+
+      if (service.pool) {
+        let pool = await this.poolRepository.findById(service.pool);
+        params.pool = pool;
+
+        if (pool.members.length > 0) {
+          let members = [];
+          for (let member_id of pool.members) {
+            let m = await this.memberRepository.findById(member_id);
```
zongzw: `find` can return the member list directly, thus, no `for of` is needed.

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +173,79 @@ export class ApplicationController {
...
+            members.push(m);
+          }
+          params.members = members;
+        } else {
+          params.members = [];
+        }
+      }
+    }
+
+    let req = new AS3DeployRequest(params);
```
zongzw: In the future, we'll make it call as3Service.deploy with async way.

app/waf/src/models/adc.model.ts
```diff
@@ -16,23 +16,30 @@ export class Adc extends Entity {
...
+    required: false,
+    default: 'admin',
   })
-  username?: string;
+  username: string;
 
   @property({
     type: 'string',
+    required: false,
+    default: 'admin',
```
zongzw: why do we need default value...
It sounds weird then user succeed without inputting any credential information.
So left it to be required. or let bigip response the unauth* message is better..

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +173,79 @@ export class ApplicationController {
...
+    } catch (e) {
+      if (e.code !== 'ENTITY_NOT_FOUND') {
+        throw e;
+      }
+    }
+
+    if (!tenantAssoc) {
+      try {
+        tenantAssoc = await this.tenantAssociationRepository.findById(
+          'default',
```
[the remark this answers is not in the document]
zhaoqin-github: Lookup associated ADCs by tenantId first. If no ADC associated, then lookup the default ADC for all tenants instead. If no default ADC, that means we are not able to deploy configuration for this application.

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +173,79 @@ export class ApplicationController {
...
+      tenantAssoc = await this.tenantAssociationRepository.findById(
+        application.tenantId,
+      );
+    } catch (e) {
+      if (e.code !== 'ENTITY_NOT_FOUND') {
+        throw e;
+      }
+    }
+
+    if (!tenantAssoc) {
```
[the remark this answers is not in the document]
zhaoqin-github: Really?

app/waf/src/controllers/application.controller.ts
zhaoqin-github: Yes, but we needn't to catch it. The framework can return 404 error to client. It works in a same way as other handlers.

app/waf/src/controllers/application.controller.ts
zhaoqin-github: Done

app/waf/src/controllers/application.controller.ts
zhaoqin-github: Agree

app/waf/src/models/adc.model.ts
zhaoqin-github: We need to modify this model, after we implement trusted group. Let it go in this way now.

app/waf/src/controllers/application.controller.ts
zhaoqin-github: Let's make that model to 'hasMany' memebers in near future.

app/waf/src/controllers/application.controller.ts
zhaoqin-github: Done

app/waf/src/controllers/application.controller.ts
zhaoqin-github: Done

app/waf/src/controllers/application.controller.ts
zongzw: `find` style of this can be: 
```
          let members = [];
          members = await this.memberRepository.find({
            where: {
              id: {
                inq: pool.members
              }
            }
          })```

File: app/waf/src/models/as3.model.ts [hunk 10]
```diff
@@ -0,0 +1,238 @@
...
+    this.targetUsername = adc.username;
+    this.targetPassphrase = adc.passphrase;
+    this.declaration = new AS3Declaration(params);
+  }
+}
+
+export class AS3Declaration extends AS3Object {
+  schemaVersion: string;
+  id: string;
+  tenants: AS3Tenant[];
```
zongzw: Why not use `tenants: Object` directly, then inline `toJSON()` function in constructor. 

So that we can leverage `JSON.stringify(<AS3 series class type>)` to convert them to string, and `JSON.parse(<json object string>)` to convert string to json object. 
The Object <-> Json process becomes uniform.

app/waf/test/acceptance/application.controller.acceptance.ts
```diff
@@ -264,4 +231,104 @@ describe('ApplicationController', () => {
...
+      });
+      let service = await givenServiceData(wafapp, {
+        pool: pool.id,
+      });
+      let application = await givenApplicationData(wafapp, {
+        services: [service.id],
+      });
+
+      let s = sinon
+        .stub(controller.as3Service, 'deploy')
```
zongzw: How about to think about how to abstract the stub framework for rest datasource if we prefer to stub it.

app/waf/src/controllers/application.controller.ts
```diff
@@ -17,18 +17,39 @@ import {
   requestBody,
   HttpErrors,
 } from '@loopback/rest';
-import {Application} from '../models';
-import {ApplicationRepository, WafpolicyRepository} from '../repositories';
+import {inject} from '@loopback/context';
```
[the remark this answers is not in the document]
zhaoqin-github: Sorry....

app/waf/src/models/as3.model.ts
```diff
@@ -0,0 +1,238 @@
...
+
+export class AS3Object extends Object {
+  class: string;
+
+  constructor() {
+    super();
+  }
+}
+
+export class AS3Pointer extends Object {
```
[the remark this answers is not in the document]
zhaoqin-github: This one will be used when declaring WAF, which will be done by Jikui's commit.

app/waf/test/acceptance/application.controller.acceptance.ts
```diff
@@ -264,4 +231,104 @@ describe('ApplicationController', () => {
...
+
+      await client
+        .post(prefix + '/applications/' + application.id + '/deploy')
+        .expect(200);
+
+      // Just intend to increase code coverage
+      let req = <AS3DeployRequest>s.getCall(0).args[2];
+      req.declaration.toJSON();
+
+      s.restore();
```
zongzw: Move to `before` and `after`.

app/waf/test/acceptance/application.controller.acceptance.ts
```diff
@@ -264,4 +231,104 @@ describe('ApplicationController', () => {
...
+
+      let s = sinon
+        .stub(controller.as3Service, 'deploy')
+        .returns(Promise.resolve('Hello'));
+
+      await client
+        .post(prefix + '/applications/' + application.id + '/deploy')
+        .expect(200);
+
+      // Just intend to increase code coverage
```
zongzw: Sounds weird.

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +173,82 @@ export class ApplicationController {
...
+      throw new HttpErrors.UnprocessableEntity(
+        'No service in Application ' + application.id,
+      );
+    }
+
+    let params: {[key: string]: Object} = {
+      application: application,
+    };
+
+    let tenantAssocs = await this.tenantAssociationRepository.find({
```
zongzw: Since we know how to build the filter, the following seems more clear. 

```
    let tenantAssocs = await this.tenantAssociationRepository.find({
      limit: 1, 
      where: {
        tenantId: {
          inq: [application.tenantId, 'default']
        },
      },
    });

    if (tenantAssocs.length === 0) {
      throw new HttpErrors.UnprocessableEntity(
        'No ADC associated with Application ' + application.id,
      );
    }
```

But, 

> Lookup associated ADCs by tenantId first. If no ADC associated, then lookup the default ADC for all tenants instead. If no default ADC, that means we are not able to deploy configuration for this application.

I don't think there should be a ADC for all tenant. User should appoint one of them for deploying, otherwise, fails it. 
What's more, 

`params.adc = adcs[0];` may be a problem. When there're multiple ADCs, we should let user to select one of them instead of hardcoding to [0].

app/waf/test/acceptance/application.controller.acceptance.ts
zhaoqin-github: I think it is already a 'ready-to-use' framework. What is your expectation?

app/waf/test/acceptance/application.controller.acceptance.ts
zhaoqin-github: deleted

app/waf/test/acceptance/application.controller.acceptance.ts
zhaoqin-github: Not all case need this.

app/waf/test/acceptance/application.controller.acceptance.ts
zongzw: Agree, restServer for restDatasource is not required. 

Here, I just want to make the `.stub(controller.as3Service, 'deploy')` reusable, which means move the stub statement to `before` and `after`.

app/waf/src/models/as3.model.ts
zhaoqin-github: The structure of AS3 JSON declaration is not consistent with object structure. Many objects, like pool, waf and etc, need to be declared in the same level with service. So we need construct the JSON structure by ourselves.

app/waf/test/acceptance/application.controller.acceptance.ts
zongzw: There are many test cases trigger it. We should move it over just like `emptyDatabase` is not required for all test cases, but it's still be encapsulated into setup/teardown.

app/waf/src/controllers/application.controller.ts
zhaoqin-github: The default ADC may be an use case. If we request every tenants to specify one ADC, that means we are attempt to import all tenants in to our db, but I do not think all of us are willing to do that. Anyway, it is not a programming question.

Since we are query ADC by its primary id, we can assume only one will be returned.

zongzw: OK for the current 1:1 model.

app/waf/src/controllers/application.controller.ts
zongzw: ok

app/waf/src/models/as3.model.ts
zongzw: got it.

app/waf/test/acceptance/application.controller.acceptance.ts
zhaoqin-github: These two parameters make stub() to be reusable. Every test case has its unique stub requirement, eg. difference class, function, stub behaviors. It does not make sense to move it to before().

zongzw: NO. 

You have several it() has the stub, they are duplicate code, encapsulation should be done. 
Move them to `before` and `after`.  Make the code as clear as possible.

zongzw: You are writing acceptance test. not unit test.
"
difference class, function, stub behaviors. " should not be the excuse of not doing encapsulation and abstraction.

app/waf/src/controllers/application.controller.ts
zhaoqin-github: Done

PR #39, opened 2019-02-28
Multiple WAF Policies

jikui:
multiple WAF Policies


app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +184,115 @@ export class ApplicationController {
...
+        params.members = members;
+      } else {
+        params.members = [];
+      }
+    }
+    let rules = [];
+    let wafs = [];
+    if (application.services.length > 0) {
+      let serviceId = application.services[0];
+      let service = await this.serviceRepository.findById(serviceId);
```
zongzw: Some of the logic is duplicate with above. Need to refactor. 
For example: `let service = await this.serviceRepository.findById(serviceId);` was done before.

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +184,115 @@ export class ApplicationController {
...
+      if (service) {
+        params.service = service;
+        let endpointpolicyId = service.endpointpolicy;
+        if (endpointpolicyId != null) {
```
zongzw: `!=` should be `!==`?

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +184,115 @@ export class ApplicationController {
...
+          let endpointpolicy = await this.endpointpolicyRepository.findById(endpointpolicyId);
+          if (endpointpolicy) {
+            params.endpointpolicy = endpointpolicy;
+            if (endpointpolicy.rules!=null) {
+              for (let ruleId of endpointpolicy.rules) {
```
zongzw: Use `find` instead of `findById` to simplify the implementation.

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +184,115 @@ export class ApplicationController {
...
+                let rule = await this.ruleRepository.findById(ruleId);
+		            if (rule) {
+                  let wafId = rule.wafpolicy;
+		              if (wafId != null && wafId != "") {
```
zongzw: Error indent.

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +184,115 @@ export class ApplicationController {
...
+              params.rules=rules;
+              params.wafs=wafs; 
+            }
+          }
+          
+        }
+      }
+    } 	
+    let req = new AS3DeployRequest(params);
+    return await this.as3Service.deploy(AS3_HOST, AS3_PORT, req);
```
zongzw: I have a question that there's `try.. catch ` for as3Service.info, why there's no of it for as3Service.deploy?

When would as3Service throw the exception? Do we really need `try .. catch `?

app/waf/src/controllers/application.controller.ts
```diff
@@ -17,18 +17,50 @@ import {
...
+  ServiceRepository,
+  PoolRepository,
+  MemberRepository,
+  WafpolicyRepository,
+  EndpointpolicyRepository,
+  RuleRepository,
+} from '../repositories';
+import {AS3Service} from '../services';
 import uuid = require('uuid');
+import { isNullOrUndefined } from 'util';
```
zongzw: import but not used?

app/waf/src/controllers/application.controller.ts
zongzw: `null` should be changed to `undefined`.

app/waf/fillin_data_new.sh
```diff
@@ -0,0 +1,138 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
jikui: will remove this file

app/waf/src/controllers/application.controller.ts
```diff
@@ -17,18 +17,50 @@ import {
...
+
+const AS3_HOST: string = process.env.AS3_HOST || '10.128.0.149';
```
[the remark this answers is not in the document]
jikui: will rollback to local host

app/waf/src/controllers/application.controller.ts
```diff
@@ -155,4 +184,115 @@ export class ApplicationController {
...
+          members.push(m);
+        }
+        params.members = members;
+      } else {
+        params.members = [];
+      }
+    }
+    let rules = [];
+    let wafs = [];
+    if (application.services.length > 0) {
```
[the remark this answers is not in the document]
jikui: due to the merge. Need to simplify the logic.

app/waf/src/controllers/application.controller.ts
jikui: due to the merge. Need to simplify the logic.

app/waf/src/controllers/application.controller.ts
jikui: corrected it.

app/waf/src/controllers/application.controller.ts
jikui: It looks well locally.

app/waf/src/controllers/application.controller.ts
jikui: resolved.

app/waf/src/controllers/application.controller.ts
zongzw: Is there some `tab` instead of `space` there because of using 'vim' as the editor?

app/waf/src/controllers/application.controller.ts
jikui: resolved

app/waf/src/controllers/application.controller.ts
```diff
@@ -250,6 +259,37 @@ export class ApplicationController {
       }
     }
 
+    let rules = [];
+    let wafs = [];
+    if (service) {
```
zongzw: This `if (service) {` can be removed.

app/waf/src/controllers/application.controller.ts
```diff
@@ -250,6 +259,37 @@ export class ApplicationController {
...
+      let endpointpolicyId = service.endpointpolicy;
+      if (endpointpolicyId !== undefined ) {
+        let endpointpolicy = await this.endpointpolicyRepository.findById(
+          endpointpolicyId,
+        );
+        if (endpointpolicy) {
+          params.endpointpolicy = endpointpolicy;
+          if (endpointpolicy.rules !== undefined) {
+            for (let ruleId of endpointpolicy.rules) {
+              let rule = await this.ruleRepository.findById(ruleId);
```
zongzw: How about to use `ruleRepository.find` to remove the for loop, what's more no exception.

app/waf/src/controllers/application.controller.ts
```diff
@@ -250,6 +259,37 @@ export class ApplicationController {
...
+    let rules = [];
+    let wafs = [];
+    if (service) {
+      params.service = service;
+      let endpointpolicyId = service.endpointpolicy;
+      if (endpointpolicyId !== undefined ) {
+        let endpointpolicy = await this.endpointpolicyRepository.findById(
+          endpointpolicyId,
+        );
+        if (endpointpolicy) {
```
zongzw: I prefer to use less if to replace `if (endpointpolicy) {`, as this: 
```
    if (service.endpointpolicy) {
      params.endpointpolicy = await this.endpointpolicyRepository.findById(
        service.endpointpolicy,
      );

      params.rules = await this.ruleRepository.find({
        where: {
          id: {
            inq: (<Endpointpolicy>params.endpointpolicy).rules,
          },
        },
      });

      params.wafs = await this.wafpolicyRepository.find({
        where: {
          id: {
            inq: (<Array<Rule>>params.rules).map(v => {
              return v.wafpolicy;
            }),
          },
        },
      });
    }
```

Reasons for this are:
1. easier to make them covered. As you know, if else is not easier to cover. 
2. use `find` is more powerful, at least no exceptions. 
3. codes look clear.

app/waf/src/controllers/application.controller.ts
jikui: corrected it.

app/waf/src/controllers/application.controller.ts
jikui: corrected it.

app/waf/src/controllers/application.controller.ts
jikui: I will try with this.

app/waf/src/datasources/as-3.datasource.json
```diff
@@ -11,20 +11,20 @@
     {
       "template": {
         "method": "GET",
-        "url": "https://{host}:{port}/mgmt/shared/appsvcs/info"
+        "url": "https://admin:admin@{host}:{port}/mgmt/shared/appsvcs/info"
```
zhaoqin-github: remove basic auth.

PR #40, opened 2019-03-04
Refactor ADC resource APIs

zhaoqin-github:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -37,15 +37,16 @@ export class AdcController {
...
-  async create(@requestBody() adc: Partial<Adc>): Promise<Adc> {
+  async create(@requestBody() adc: Partial<Adc>): Promise<AdcResponse> {
     if (!adc.id) {
       adc.id = uuid();
     }
 
     try {
-      return await this.adcRepository.create(adc);
+      let data = await this.adcRepository.create(adc);
+      return new AdcResponse(data);
```
zongzw: All `Promise<xxxx>` functions should return `Promise.resolve(...)` or `Promise.reject(...)`.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -37,15 +37,16 @@ export class AdcController {
...
   })
-  async create(@requestBody() adc: Partial<Adc>): Promise<Adc> {
+  async create(@requestBody() adc: Partial<Adc>): Promise<AdcResponse> {
     if (!adc.id) {
       adc.id = uuid();
     }
 
     try {
-      return await this.adcRepository.create(adc);
+      let data = await this.adcRepository.create(adc);
```
zongzw: The `data` variable is only used once after the declaration/assignment. 
I'm thinking that if we need to remove it, thus: 

(not check whether or not can run)
```
return new AdcResponse(
    await this.adcRepository.create(adc)
);

```
Only one statement...

zongzw: It's not right or wrong, but only concern about "Code of Conduct" or standard.

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: Done

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: Done

app/waf/src/controllers/common.controller.ts
```diff
@@ -0,0 +1,296 @@
+import {ParameterObject, ParameterLocation} from '@loopback/openapi-v3-types';
+
+const adcCreateRequestProperties = {
+  id: {
+    type: 'string',
+  },
+  name: {
```
zongzw: Defining the whole bunch of things is boring since they are very similar to our model definition.
I'm wondering if we can generate them from our model definition, typescript support reflection. 
How about using the following possible code core to do the schema definition: 

```  
let props = Adc.definition.properties;
for(let p of ['id', 'name', 'createdAt', 'UpdatedAt']) {
    delete props[p];
}

adcCreateRequestProperties = props;
```

app/waf/src/controllers/common.controller.ts
```diff
@@ -0,0 +1,296 @@
+import {ParameterObject, ParameterLocation} from '@loopback/openapi-v3-types';
+
+const adcCreateRequestProperties = {
```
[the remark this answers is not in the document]
zhaoqin-github: We have adcUpdateRequestProperties and other request bodies whose property lists are different. So we name them different.

app/waf/src/controllers/common.controller.ts
zhaoqin-github: Let me think

app/waf/src/controllers/common.controller.ts
```diff
@@ -0,0 +1,296 @@
+import {ParameterObject, ParameterLocation} from '@loopback/openapi-v3-types';
```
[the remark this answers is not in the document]
zhaoqin-github: Done.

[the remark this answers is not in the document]
zhaoqin-github: Since it is api schema, so we would save it under controller directory.

app/waf/src/controllers/common.controller.ts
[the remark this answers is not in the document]
zongzw: Is there any different between create and update? suppose they are using the same type(class) with only different set of properties.

How about to use `name?: ..` to make the properties selectable?

app/waf/src/controllers/common.controller.ts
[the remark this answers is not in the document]
zongzw: The controllers folder is in special usage, see application.ts: 

```
    this.bootOptions = {
      controllers: {
        // Customize ControllerBooter Conventions here
        dirs: ['controllers'],
        extensions: ['.controller.js'],
        nested: true,
      },
    };
```

I think we'd better to put only real controllers under it.

app/waf/src/controllers/common.controller.ts
zhaoqin-github: Done.

app/waf/src/controllers/common.controller.ts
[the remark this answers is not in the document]
zhaoqin-github: The update body will allow less properties. For example, 'id' property is not allowed in update request body.

This schema is used to generate API docs. Required properties are specified by 'required' list. Optional properties doe not require to be specified.

app/waf/src/controllers/common.controller.ts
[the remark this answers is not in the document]
zhaoqin-github: Those variables are for controllers to initialize, so it does make sense to place them under controller directory.

app/waf/src/controllers/schema.ts
```diff
@@ -0,0 +1,284 @@
...
+  passphrase: 'admin',
+};
+
+const adcUpdateRequestProperties = buildProperties(adcProperties, ['name']);
+
+const adcUpdateRequestExample = {
+  name: 'My BIG-IP',
+};
+
+const adcResponseProperties = buildProperties(adcProperties, [
```
zongzw: Do we have to list the properties, can we find them from model definition?

app/waf/src/controllers/schema.ts
```diff
@@ -0,0 +1,284 @@
...
+          required: requiredProperties,
+          properties: properties,
+          example: example,
+        },
+      },
+    },
+  };
+}
+
+function buildResponseSchema(properties: Object, example: Object): Object {
```
zongzw: `Object` is not a internal type, why not use `object`?

app/waf/src/controllers/adc.controller.ts
```diff
@@ -32,27 +31,35 @@ export class AdcController {
...
-    if (!adc.id) {
-      adc.id = uuid();
-    }
-
+  async create(
+    @requestBody(Schema.adcCreateRequest) reqBody: Partial<Adc>,
+  ): Promise<AdcResponse> {
     try {
-      return await this.adcRepository.create(adc);
+      return Promise.resolve(
```
zongzw: My fault that, here, returning Promise and value itself are both OK. 
But in catch block, I suggest `throw new HttpErrors.BadRequest(error.message);`

app/waf/src/controllers/schema.ts
```diff
@@ -0,0 +1,284 @@
...
+      items: {
+        type: 'object',
+        properties: properties,
+        example: example,
+      },
+    },
+  };
+}
+
+export class Schema {
```
zongzw: Will we rename the class name from `Schema` to 'AdcSchema', since there may be more schema for all models?

app/waf/src/controllers/schema.ts
```diff
@@ -0,0 +1,284 @@
...
+      props[propName] = {
+        type: metadata[propName].type,
+      };
+    }
+  }
+
+  return props;
+}
+
+const adcProperties = MetadataInspector.getAllPropertyMetadata<
```
zongzw: I'm not familiar with MetadataInspector. 

In fact, there may be possible thinking, may be helpful to you to make the implementation easier. 

In our model definition, it supports define our own meta properties, like this: 
```
// ... from application.model.ts
  @property({
    type: 'array',
    itemType: 'string',
    required: false,
    default: [],
    inResp: true,
    inReq: false
  })
  services: string[];
```

```
console.log(JSON.stringify(Application.definition.properties));
// output
... "services":{"type":"array","itemType":"string","required":false,"default":[],"inResp":true,"inReq":false}, ...
```

I'm very concern about this point. because: 
It would be a big workload if don't design carefully on the documentation process...
Can we make it possible to cut down ANY duplicate code pieces as possible as we can?

app/waf/src/controllers/common.controller.ts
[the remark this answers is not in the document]
zongzw: That's ok to place here if you think ok. 
But just concern in the future, if the schema becomes bigger or more, we'd better put them together, be well cataloged.

app/waf/src/controllers/schema.ts
zhaoqin-github: OK

app/waf/src/controllers/schema.ts
zhaoqin-github: We can attempt to define our own decorator to do that, but that is not containable in this pull request. Suggest to open another branch to do that.

app/waf/src/controllers/schema.ts
zhaoqin-github: Done

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: Let us change them from Promise.resolve/reject back to 'return' and 'throw'.

zhaoqin-github: Done.

app/waf/src/controllers/schema.ts
zhaoqin-github: Done.

zongzw:
It's OK for now, could you please add a 'TODO', so that we can easily get all future tasks.

zhaoqin-github:
@zongzw Done

PR #41, opened 2019-03-04
Pool has many members

zhang-shengping:


app/waf/.vscode/launch.json
```diff
@@ -1,14 +0,0 @@
-{
-  "version": "0.2.0",
```
zongzw: Why do we remove the .vscode file?

app/waf/src/controllers/application.controller.ts
```diff
@@ -247,12 +247,8 @@ export class ApplicationController {
...
       params.pool = pool;
 
-      if (pool.members.length > 0) {
-        let members = [];
-        for (let member_id of pool.members) {
-          let m = await this.memberRepository.findById(member_id);
-          members.push(m);
-        }
+      let members = await this.poolRepository.members(pool.id).find();
+      if (members.length > 0) {
```
zongzw: No need to check length??

```
params.members = await this.poolRepository.members(pool.id).find();
```

app/waf/src/controllers/pool.controller.ts
```diff
@@ -238,43 +239,25 @@ export class PoolController {
...
-        members.push(await this.memberRepository.findById(mbr));
-      }
-      return members;
-    } else {
-      return [] as Member[];
-    }
+    @param.path.string('member_id') member_id: string,
+    @requestBody() member: Partial<Member>,
+  ): Promise<Count> {
+    const result = await this.poolRepository
```
zongzw: The `result` variable is only used once. Could we remove the declaration?

PR #42, opened 2019-03-07
App has many sv

zhang-shengping:


app/waf/src/models/application.model.ts
```diff
@@ -25,13 +26,10 @@ export class Application extends Entity {
...
   tenantId: string;
-  @property({
-    type: 'array',
-    itemType: 'string',
-    required: false,
-    default: [],
-  })
-  services: string[];
+
+  @hasMany(() => Service, {keyTo: 'appId'})
```
zhaoqin-github: appId --> applicationId  ?

app/waf/src/models/service.model.ts
```diff
@@ -40,6 +40,12 @@ export class Service extends Entity {
   })
   endpointpolicy: string;
 
+  @property({
+    type: 'string',
+    required: true,
+  })
+  appId: string;
```
zhaoqin-github: appId --> applicationId ?

This property name can be exposed by REST API.

app/waf/test/acceptance/service.controller.acceptance.ts
```diff
@@ -9,8 +9,9 @@ import {setupApplication, teardownApplication} from '../helpers/test-helper';
 import {
   givenEmptyDatabase,
   givenServiceData,
-  createServiceObjectWithoutID,
+  createServiceObject,
```
zhaoqin-github: Currently our API allows to input resource uuid. I suggest we can continue to have this behavior, until we confirm that it is not require any more. Please continue to have two test cases, one input uuid, one does not.

app/waf/tsconfig.json
```diff
@@ -3,7 +3,8 @@
     "module": "commonjs",
     "target": "es5",
     "outDir": "./dist",
-    "sourceMap": true
+    "sourceMap": true,
+    "experimentalDecorators": true
```
zhaoqin-github: Why we needn't this when we implement hasMany for pool/member?

zongzw: same question. why do we need it? The code seems run/test well without it.

app/waf/test/acceptance/service.controller.acceptance.ts
zongzw: In fact, I think we can combine the `createServiceObjectWithoutID` into `createServiceObject` by passing uuid: 
```
createServiceObject({
    'id': uuid()
});
```
No need to declare a new function with `WithoutID` or `WithID`.

app/waf/tsconfig.json
zhang-shengping: it will show some experimental decorator error on my VS, if not add these

app/waf/src/controllers/application.controller.ts
```diff
@@ -238,8 +247,7 @@ export class ApplicationController {
     }
     params.adc = adcs[0];
 
-    let serviceId = application.services[0];
-    let service = await this.serviceRepository.findById(serviceId);
+    let service = <Service>services[0];
```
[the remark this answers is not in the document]
zhang-shengping: no, it may undefined

app/waf/test/acceptance/service.controller.acceptance.ts
zhang-shengping: agree to change it

app/waf/src/controllers/application.controller.ts
```diff
@@ -238,8 +247,7 @@ export class ApplicationController {
...
 
     params.service = service;
```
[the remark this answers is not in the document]
zhang-shengping: you can not use this "params.service" to get service object attributes like this way "params.service.id", thus I kept service variable.

app/waf/src/models/service.model.ts
zhang-shengping: I will change it in next push

app/waf/src/controllers/application.controller.ts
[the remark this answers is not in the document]
zongzw: `it`, do you mean `services` ?

app/waf/src/controllers/application.controller.ts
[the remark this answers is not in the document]
zongzw: I mean there are two lines of `params.service = service;`, is there?

app/waf/tsconfig.json
zongzw: I tried `npm start` `npm test` `npm run build`. There's no error happens.

app/waf/test/acceptance/service.controller.acceptance.ts
zhang-shengping: I have changed the code as the way Andrew suggested.

PR #44, opened 2019-03-08
Publish API specification

zhaoqin-github:


.travis.yml
```diff
@@ -3,12 +3,42 @@ language: node_js
...
+            for file in swagger-ui.css swagger-ui-bundle.js swagger-ui-standalone-preset.js favicon-16x16.png favicon-32x32.png ; do
+              curl -s -X GET http://${HOST}/explorer/${file} > ${file}
+            done
+          - sed -i 's/LoopBack API Explorer/F5 ADCaaS API Specification/g' index.html
+          - sed -i 's/\/openapi.json/.\/openapi.json/g' index.html
+          - git config --global user.email "q.zhao@f5.com"
+          - git config --global user.name "Qin Zhao"
+          - git config --global credential.helper "$(pwd)/git-credential.sh"
+          - git commit -a -m "Travis CI $TRAVIS_BUILD_WEB_URL"
+          - git push origin gh-pages
```
zhang-shengping: Does it use your personal own git pages?

.travis.yml
```diff
@@ -3,12 +3,42 @@ language: node_js
...
+          - git clone https://github.com/F5Networks/f5-adcaas-openstack.git
+          - cd f5-adcaas-openstack
+          - git checkout gh-pages
+          - HOST="localhost:3000"
+          - curl -s -X GET http://${HOST}/openapi.json > openapi.json
+          - curl -s -X GET http://${HOST}/explorer/ > index.html
+          - |
+            for file in swagger-ui.css swagger-ui-bundle.js swagger-ui-standalone-preset.js favicon-16x16.png favicon-32x32.png ; do
+              curl -s -X GET http://${HOST}/explorer/${file} > ${file}
+            done
```
zongzw: What's the benefit to write each line with '-' instead of encapsulating them into a script?

If you want the command lines to be logged, you can use `set -x`; 
If you want to quit immediately on error, you can use `set -e`

I really disagree t with them to list here, which is not suitable for the run/test locally before `git commit`. 
Just as I disagree that in our pipeline did.

.travis.yml
zhaoqin-github: No. It is the github pages of F5 account. I need to make a commit message here.

.travis.yml
```diff
@@ -3,12 +3,42 @@ language: node_js
...
       script:
         - npm --prefix app/waf test
         - npm --prefix app/waf run coverage
         - npm --prefix app/waf run coveralls
+      deploy:
+        provider: script
+        on:
+          branch: master
+        script:
+          - npm --prefix app/waf start >/dev/null 2>&1 &
```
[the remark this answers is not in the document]
zhaoqin-github: May we can. Let me have a try.

.travis.yml
zhaoqin-github: Let us reconsider it, after removing the multi-line script here.

.travis.yml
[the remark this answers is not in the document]
zhaoqin-github: CSS and JS files are almost same, but LB4 rebrand the HTML file. And LB4 does not support to generate openapi.json currently, so we need to get index.html and openapi.json at least.

.travis.yml
zhaoqin-github: Move them to a script now.

PR #47, opened 2019-03-12
add monitor resource

zhang-shengping:


app/waf/src/controllers/monitor.controller.ts
```diff
@@ -0,0 +1,141 @@
...
+      },
+    },
+  })
+  async find(
+    @param.query.object('filter', getFilterSchemaFor(Monitor)) filter?: Filter,
+  ): Promise<Monitor[]> {
+    return await this.monitorRepository.find(filter);
+  }
+
+  @patch(prefix + '/monitors', {
```
zhaoqin-github: Please remove "updateAll". This API is not required.

app/waf/src/controllers/monitor.controller.ts
```diff
@@ -0,0 +1,141 @@
...
+    },
+  })
+  async updateById(
+    @param.path.string('id') id: string,
+    @requestBody() monitor: Monitor,
+  ): Promise<void> {
+    await this.monitorRepository.updateById(id, monitor);
+  }
+
+  @put(prefix + '/monitors/{id}', {
```
zhaoqin-github: Please remove PUT API.

app/waf/src/models/monitor.model.ts
```diff
@@ -0,0 +1,46 @@
+import {Entity, model, property} from '@loopback/repository';
+
+@model()
```
zhaoqin-github: Suggest to include all possible monitor properties, so that we will have all columns defined in db. It will be easier to support new types of monitor in the future.

app/waf/src/models/monitor.model.ts
```diff
@@ -0,0 +1,46 @@
...
+    id: true,
+    required: true,
+  })
+  id: string;
+
+  @property({
+    type: 'number',
+    default: 5,
+  })
+  interval?: number;
```
zhaoqin-github: If a property always have a default value, we may remove '?', so that the var type will be a little simpler.

app/waf/src/models/monitor.model.ts
```diff
@@ -0,0 +1,46 @@
+import {Entity, model, property} from '@loopback/repository';
+
+@model()
+export class Monitor extends Entity {
```
zhaoqin-github: Suggest to inherit CommonEntity in models/common.model.ts, who helps to define id, name, createdAt and updatedAt properties.

app/waf/src/models/monitor.model.ts
```diff
@@ -0,0 +1,287 @@
+import {model, property} from '@loopback/repository';
+import {CommonEntity} from './common.model';
+
+@model()
+export class Monitor extends CommonEntity {
```
zongzw: Putting different type of monitor attributes into one model is not a reasonable way, but should work for now. 

I think we can investigate the polymorphism which is to define multiple models for the subtype of the monitor separately. And use the parent monitor type as other model's reference. 

Could you please add a TODO item to mark this task so that you can commit the code right now with currently implemented way?

zongzw: I do some test on the inheritance of model, also the usage in repository/controller. Separate each type of monitors into different model definition should be good. 

https://github.com/zongzw/learn-typescript/tree/master/src/99.verifications/lb4/inheritance-test

zhang-shengping: I can write that as you said, but right now I will switch to lbaasv2, setup env, learn how to write irules and investigate l7policy code. We will discuss this later.

zongzw: Good and ok for that. We'll do it in the future.

app/waf/src/models/monitor.model.ts
```diff
@@ -0,0 +1,287 @@
...
+  @property({
```
zhaoqin-github: Please remove id property. It is defined in CommonEntity.

app/waf/src/models/monitor.model.ts
```diff
@@ -0,0 +1,287 @@
...
+  })
+  chaseReferrals: boolean;
+
+  @property({
+    type: 'string',
+    default: '',
+  })
+  ciphers: string;
+
+  @property({
```
zhaoqin-github: Please remove class property. It is only for AS3 JSON, and will be filled during translating.

PR #48, opened 2019-03-14
Openstack integration test

zongzw:


zongzw:
Coverage is back to > 90%, more coverage would be done during/after the refactoring.

app/waf/src/application.ts
```diff
@@ -51,5 +56,61 @@ export class WafApplication extends BootMixin(
         nested: true,
       },
     };
+
+    if (process.env.TEST_OPENSTACK_INTEGRATION)
```
zhaoqin-github: Can you briefly describe why we just behavior like this only for "TEST_OPENSTACK_INTEGRATION"?

app/waf/.vscode/launch.json
```diff
@@ -1,14 +1,46 @@
...
       "type": "node",
       "request": "launch",
       "program": "${workspaceFolder}/index.js",
       "args": [],
       "cwd": "${workspaceFolder}",
-      "protocol": "inspector"
+      "protocol": "inspector",
+      "env": {
+        "TEST_OPENSTACK_INTEGRATION": "1"
+        // "PRODUCT_RELEASE": "1",
```
zhaoqin-github: Now I prefer to remove .vscode from repo, after looking this... since it benefit very little to other contributors. We should encourage the diversity among contributors. Everyone can manage their ow editor and configuration file by themself, instead of mange them in a shared repo.

app/waf/src/index.ts
```diff
@@ -8,6 +8,7 @@ const prefix = '/adcaas/v1';
 export async function main(options: ApplicationConfig = {}) {
   const app = new WafApplication(options);
   await app.boot();
+  app.keepAliveAdminToken();
```
zhaoqin-github: Is it a background job? Or only run once?

app/waf/src/application.ts
```diff
@@ -51,5 +56,61 @@ export class WafApplication extends BootMixin(
...
+      this.controller(OpenstackController);
+  }
+
+  async keepAliveAdminToken() {
+    if (!process.env.PRODUCT_RELEASE) return;
```
zhaoqin-github: What does it mean?

app/waf/src/sequence.ts
```diff
@@ -41,7 +49,7 @@ export class MySequence implements SequenceHandler {
       this.reject(context, err);
     }
 
-    // await this.logResponse(logUuid, context, result);
+    await this.logResponse(logUuid, context, result);
```
zhaoqin-github: Did we fix the bug of response log?

app/waf/src/sequence.ts
```diff
@@ -32,6 +37,9 @@ export class MySequence implements SequenceHandler {
     let result: object = {};
     try {
       const {request, response} = context;
+
+      //await this.authRequest(request);
```
zhaoqin-github: Still need this?

app/waf/.vscode/launch.json
zongzw: This may draw another topic: I recommend that we use the very same developing environment. VS/VSCode is the recommended IDE from MS for typescript.
However, not forced. It's OK to remove it from the code repo, personally, I'll keep it myself. 


It's OK to remove it from the code repo, personally, I'll keep it myself.

app/waf/src/application.ts
zongzw: In product releaese, we don't need to expose openstack operation API as demo shows. 
So this is only for demo. Will remove it.

app/waf/src/application.ts
zongzw: I don't want the auth process to bother us during the developing. So this is an on-off if we really need to connect OpenStack for auth*.

app/waf/src/index.ts
zongzw: A background job. You can review the implementation of keepAliveAdminToken, it will keep the auth token active by loopy logging in again before it expires.

app/waf/src/sequence.ts
zongzw: Yes. If we decide to integrate with OpenStack, we have to auth every user's request before handling it.

zongzw: It should also be wrapped with `if (PRODUCT_RELEASE) ...`, I think.

app/waf/src/sequence.ts
zongzw: Not yet. commented on it and Add a TODO.

app/waf/src/application.ts
```diff
@@ -51,5 +56,61 @@ export class WafApplication extends BootMixin(
...
+      'services.openstack.AuthWithOSIdentity',
+    );
+
+    let delayFunc = async (ms: number) => {
+      return new Promise(resolve => setTimeout(resolve, ms).unref());
+    };
+
+    while (true) {
+      if (tried > totalTris) {
+        throw new Error(
```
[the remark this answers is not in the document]
zongzw: When an exception happens, the process will quit the loop.

app/waf/src/application.ts
```diff
@@ -51,5 +56,61 @@ export class WafApplication extends BootMixin(
...
+          `Tried ${tried} times connecting identity, but all fails.`,
+        );
+      }
+
+      await delayFunc(durInMillSecs);
+
+      try {
+        await authWithOSIdentity.adminAuthToken().then(
```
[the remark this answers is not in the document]
zongzw: Will Neutron client have also the same case? I'm afraid that would lead many auth code within each service...

app/waf/src/models/application.model.ts
```diff
@@ -26,7 +26,7 @@ export class Application extends Entity {
     required: true,
   })
   tenantId: string;
-
+
```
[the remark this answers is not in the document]
zongzw: In fact, not.

app/waf/src/sequence.ts
```diff
@@ -77,4 +84,28 @@ export class MySequence implements SequenceHandler {
...
+    let userToken = request.header('X-Auth-Token');
+    if (typeof userToken !== typeof '') {
+      throw new HttpErrors.Unauthorized(
+        'Unauthorized: invalid X-Auth-Token header.',
+      );
+    }
+
+    await this.authWithOSIdentity.validateUserToken(<string>userToken).then(
+      authedObj => {
+        // ...
```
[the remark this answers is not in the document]
zongzw: It should be saved to Request context, however, for now, this part is not implemented yet.
// ... should be the place for this saving job.

app/waf/src/models/adc.model.ts
```diff
@@ -6,33 +6,60 @@ export class Adc extends CommonEntity {
   @property({
     type: 'string',
     required: true,
+    schema: {
```
zhaoqin-github: Not sure why you have this diff... These code is already in master branch.

app/waf/src/models/common.model.ts
```diff
@@ -7,24 +7,42 @@ export abstract class CommonEntity extends Entity {
     type: 'string',
     id: true,
     required: true,
+    schema: {
```
zhaoqin-github: Not sure why you have this diff... These code is already in master branch.

app/waf/src/controllers/schema.ts
```diff
@@ -1,181 +1,71 @@
...
-  if (metadata) {
-    for (let propName of propNames) {
-      props[propName] = {
-        type: metadata[propName].type,
-      };
-    }
-  }
-
-  return props;
+import {PropertyDefinition} from '@loopback/repository';
```
zhaoqin-github: I think you may look into your branch right now...

PR #52, opened 2019-03-15
Implement many-to-many relation between ADC and Tenant

zhaoqin-github:


app/waf/src/controllers/schema.ts
```diff
@@ -246,9 +256,30 @@ function buildCollectionResponseSchema(
   };
 }
 
+export class TenantSchema {
+  static id = buildParameterSchema(
```
zongzw: 27 // TODO: It is betther to have another decorator or something else
 28 // to specify the resource properties of request/response in model
 29 // definition, instead of re-define them here. Need to refactor
 30 // the code in the future.


As you mentioned, there is some known better way to do it, why not try it now. If we don't build the base now, it would become difficult to refactor in the future.

zhaoqin-github: This is no capacity to contain it in this sprint. Need to complete planned tasks at first.

zhaoqin-github: A 'good' design is not free, it required time to make the design. This pull request is to implement many-to-many relation, instead of refactor schema generating code. I can have a quick try with the existing decorate, in order to find out if we can make it a little better. If we cannot, let us consider that refactoring work in another thread, not in this PR. An open code review process is good, but it is not very good to freely stretch the scope during code review.

zhaoqin-github: Please look at #55

app/waf/src/models/tenantassoc.model.ts
```diff
@@ -1,31 +1,53 @@
...
-    type: 'date',
-  })
-  createdAt?: string;
+  constructor(data?: Partial<AdcTenantAssociation>) {
+    super(data);
+  }
+}
 
+@model()
+export class Tenant extends Entity {
```
zongzw: To my understanding, we will not store any things about tenant except an Id. 
Do you really need the model?

zhaoqin-github: model decorator does not impact anything. Let's remove it.

zongzw: My concern is do we need the Tenant class?

zhaoqin-github: Just hope to generate the API doc and response in a consistent way with others. If we do not define its property and schema, it will become hard to generate api doc at least.

zongzw: I mean there's no need to have the class `Tenant`. Because we only retain a tenant Id, which is only a string in other tables.

zhaoqin-github: Yes, but we need a class with property decorator to generate API doc. Can you please propose your idea?

zongzw: Tenant(the Id, of string type) is only a field of other tables, instead of a model/table, I think.

zhaoqin-github: Maybe you did not get what I mean...

PR #53, opened 2019-03-18
conform to the naming conversion and reuse the commonEntity base class

jikui:
conform to the naming conversion and reuse the commonEntity base class

### DIFF
--- a/app/waf/test/acceptance/action.controller.acceptance.ts
+++ b/app/waf/test/acceptance/action.controller.acceptance.ts
@@ -1,4 +1,4 @@
-import {Client, expect, toJSON} from '@loopback/testlab';
+import {Client, expect} from '@loopback/testlab';
 import {WafApplication} from '../..';
 import {setupApplication, teardownApplication} from '../helpers/test-helper';
 import {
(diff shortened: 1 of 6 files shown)


app/waf/test/acceptance/action.controller.acceptance.ts
```diff
@@ -42,22 +42,13 @@ describe('ActionController', () => {
...
-
-  it('get ' + prefix + '/actions', async () => {
-    const action = await givenActionData(wafapp, {id: uuid(), ruleId: uuid()});
-
-    await client
-      .get(prefix + '/actions')
-      .query({filter: {where: {ruleId: action.ruleId}}})
-      .expect(200, [toJSON(action)]);
+      .query({where: {ruleId: action.ruleId}})
+      .expect(200);
```
zongzw: `expect(response.body).to.containDeep({id: xxxxxx});`

it's better to check the content of the response. Only part of the response.

However, the best way to find out the root cause together with @zhaoqin-github why and where the time does set.

jikui: resolved

PR #55, opened 2019-03-20
Refactor API schema generating code

zhaoqin-github:


app/waf/src/controllers/schema.ts
```diff
@@ -258,26 +226,28 @@ export class AdcSchema {
 
   static adcCreateRequest = buildRequestSchema(
     'ADC resource that need to be created',
-    ['type', 'host'],
-    adcCreateRequestProperties,
-    adcCreateRequestExample,
+    adcProperties.required,
```
zongzw: The aim of refactoring is removing ANY extra works of doc for a specific model. 
So I expect to remove any references/codes for a specific model, like here `adc`. 
Or else, we have to add more when 'doc' more other models, like `Application` `Endpointpolicy` ... 

What I imagine the implementation could be: 

```
  @post(prefix + '/adcs', {
    responses: {
      '200': {
        description: 'Successfully create ADC resource',
        content: {'application/json': Schema.getResponse(Adc)},
      },
      '400': {
        description: 'Invalid ADC resource',
        content: {'application/json': Schema.getBadRequest(Adc)},
      },
      '422': {
        description: 'Unprocessable ADC resource',
        content: {'application/json': Schema.getUnprocessableEntity(Adc)},
      },
    },
  })
  async create(
    @requestBody(Schema.getCreateRequest(Adc)) reqBody: Partial<Adc>,
  ): Promise<AdcResponse> {
    try {
...
```

What's more, I hope you can move the schema.ts out of controllers folder, it's not a controller. 
It's not necessary to put them together for referring. Make the code as clean as possible.

zhaoqin-github: Let me have a try to make it a little more simpler.

And I think currently there is not any other better place to store schema.ts. Since it is part of controller initializing facilities, it does make sense to place it under controller folder.

zongzw: How about putting it under `src` directly, it's actually not related with controller but document. Just like `log4ts.ts` `components.ts` `keys.ts` ... In the future if we think there are too many .ts under `src`, we can categorize them then, which is not a big job.

zhaoqin-github: Let us focus on other constructive topics.

zongzw: Just wonder that is there any special reason that you stick on keeping the file in controllers?

app/waf/src/controllers/schema.ts
```diff
@@ -1,181 +1,67 @@
...
-  if (metadata) {
-    for (let propName of propNames) {
-      props[propName] = {
-        type: metadata[propName].type,
-      };
-    }
-  }
-
-  return props;
+class SchemaProperties {
```
zongzw: There's no need to keep this `SchemaProperties`. 

Very simple way: Find respective parts from `propertyMetadata`, return object directly. 

Your code is time cost, for each Schema.xxxx(), it has to calculate all `create` `update`, `response`... .

app/waf/src/controllers/adc.controller.ts
```diff
@@ -28,24 +28,17 @@ export class AdcController {
...
-      },
-      '400': {
-        description: 'Invalid ADC resource',
-        content: {'application/json': AdcSchema.BadRequest},
-      },
-      '422': {
-        description: 'Unprocessable ADC resource',
-        content: {'application/json': AdcSchema.UnprocessableEntity},
-      },
+      '200': Schema.response(Adc, 'adc', 'Successfully create ADC resource'),
```
zongzw: There's no cases till now the key name in response is NOT `entity.modelName`.  case insensitive 

Could you please remove the parameter, and get it within `response` function?

app/waf/src/controllers/schema.ts
```diff
@@ -220,69 +106,180 @@ function buildRequestSchema(
...
       type: 'object',
-      properties: properties,
-      example: example,
+      properties: resp,
     },
   };
 }
 
 function buildCollectionResponseSchema(
+  name: string,
```
zongzw: Just feel that the usage of this function is little. return that object is more direct and clear.

zhaoqin-github: Why little? Every collection response api need this schema.

app/waf/src/controllers/schema.ts
zhaoqin-github: Done

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: Done

zhaoqin-github: Done

app/waf/src/controllers/schema.ts
zongzw: 1. The function is called only once.
2. What does the function do?
Only returns a wrapped object with `schema`. How about to re-define collectionResponse function with: 

```
  static collectionResponse(
    entity: typeof Entity,
    name: string,
    desc: string,
  ): object {
    let props = buildProperties(propertyMetadata(entity));
    return {
      description: desc,
      content: {
        'application/json': {
          schema: {
            type: 'object',
            properties: {
              [name]: {
                type: 'array',
                items: {
                  type: 'object',
                  properties: props.response,
                  example: props.responseExample,
                },
              }
            },
          },
        },
      },
    };
  }
```

That's is clear for us to see the whole body of the schema definition.

zongzw: I just think there's no need to define a separate function for constructing the schema, which it makes the schema not so human readable.

app/waf/src/controllers/schema.ts
zhaoqin-github: I change the latest implementation back to the original idea, that is to parse property metadta once and save it.

app/waf/src/controllers/schema.ts
zhaoqin-github: That should be optional. It seem not very worth to doing this right way.

zongzw: Make it clear.

PR #58, opened 2019-03-23
Add starting scripts and docker-compose settting for DO and ASG.

zongzw:


deploy/docker-compose.yml
```diff
@@ -27,12 +30,28 @@ services:
     volumes:
       - ${DATABASE_DATA_DIRECTORY:?err}:/var/lib/postgresql/data
 
-  AS3:
-    image: f5devcentral/f5-as3-container:3.7.0
-    container_name: as3-server
+  ASG:
```
zhaoqin-github: I see f5-api-services-gateway is the base container image, which does not have AS3 installed. Will you plan to have AS3 and ASG container installed both? Or to install other extension rpms in AS3 container?

zongzw: We may need only one of them. From Gruber's demo, it uses ASG with RPMs installed, the rpm's includes not only as3 extension but also TrustedASMpolicies/TrustedDevices/TrustedProxy. 
Not sure which option to use, however, Gruber showed us a successful option.

zhaoqin-github: But I did not see AS3 rpm in your code...

zongzw: In the demo, it uses TrustedProxy to deploy AS3 declaration. 
You know well  AS3 setup, do you know the relationship between AS3 and TrustedProxy?

zhaoqin-github: In the demo meeting, the AS3 rpm is installed by TrustedProxy on VE instance.  The topology is: client --> ASG container (with TrustedProxy installed) --> VE (with AS3 rpm installed).

Now our WAF app topology is client --> WAF API --> AS3 container --> VE (without AS3 rpm installed).

Install AS3 on VE should be a good way to improve performance, and it requires our app to switch the southbound API call  from AS3 API to TrustedProxy.

zhaoqin-github: And in the demo, the AS3 rpm seems to be downloaded by VE instance from internet directly. I am not quite sure now, if we can really push the AS3 rpm from ASG container to VE instance. Need to check that.

zongzw: Both ways should be OK. I have no idea which one to use, too. 
So let's left them both here with one commented but keep the ports same(8443 and 8080) in case we can switch them easily. 

Personally speaking, I prefer as3-on-ve, because of the bug reported last time(3.7.0 not works well unless it is installed to ve.).

PR #66, opened 2019-03-27
Refactor Service resource API

zhaoqin-github:


app/waf/src/models/service.model.ts
```diff
@@ -1,50 +1,344 @@
...
+    },
+  })
+  applicationId: string;
+
+  @property({
+    type: 'boolean',
+    required: false,
+    default: true,
+  })
+  addressStatus: boolean;
```
zongzw: Will add the schemas for these properties in the future?

zhaoqin-github: The properties without schema definition is not supported by our API now. They are excluded from response and will be disabled by the validator (todo). Will add schema, once we support those properties.

zongzw: ok

PR #74, opened 2019-03-28
Another implement of API response

zhaoqin-github:


app/waf/src/controllers/common.ts
```diff
@@ -0,0 +1,17 @@
...
+  constructor(t: typeof Entity, data: Entity) {
+    this[t.modelName.toLowerCase()] = data;
+  }
+}
+
+export class CollectionResponse {
+  [key: string]: Entity[];
+
+  constructor(t: typeof Entity, data: Entity[]) {
+    this[t.modelName.toLowerCase() + 's'] = data;
```
zongzw: https://www.npmjs.com/search?q=plural

zhaoqin-github: It is an interesting lib, but it will eventually call " + 's' ", too. It seems no value to utilize it.

zhaoqin-github: Done

PR #80, opened 2019-04-01
Build resource reponse body

zhaoqin-github:


app/waf/src/controllers/common.ts
```diff
@@ -1,19 +1,37 @@
-import {Entity} from '@loopback/repository';
+import {Entity, AnyObject} from '@loopback/repository';
```
zongzw: Just noticed that you named the file to common.ts. 

Why should we separate `Response` `CollectionResponse` to it. 
What's the benefit to put a new file to keep them? How about putting them into schema.ts?

Personally speaking, I don't like the name `common`, it let me know nothing. I suggest you to rename the following files name: 

controllers/common.ts => remove it. 
models/common.model.ts => model.base.ts or base.model.ts
repositories/common.ts => model.base.ts

zhaoqin-github: Since the 'Response' and 'CollectionResponse' class has different input and output, I create them separately.  They are not schemas, so I did not add them into that file. The file name is not very important for this program,  it should be fine to renaming the file.

zhaoqin-github: The 'common' is usually used, just because it means nothing, so that anything can be placed into a 'common' file.

zhaoqin-github: Done

PR #82, opened 2019-04-02
Retrieve tenant id from the request context.

zongzw:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -31,6 +36,9 @@ export class AdcController {
     public adcRepository: AdcRepository,
     @repository(AdcTenantAssociationRepository)
     public adcTenantAssociationRepository: AdcTenantAssociationRepository,
+    @inject(RestBindings.Http.CONTEXT)
+    private reqCxt: RequestContext,
```
zhaoqin-github: If we associate request context with controller, how do we switch the context for different request? Is it one context instance, or different instances?

zongzw: The controller instance is created for each request(request context).

zhaoqin-github: OK, got it.

app/waf/src/sequence.ts
```diff
@@ -95,22 +97,32 @@ export class MySequence implements SequenceHandler {
...
       );
+
+    if (request.headers['tenant-id']) {
+      authedToken.tenantId = <string>request.headers['tenant-id'];
+    }
+    if (!authedToken.tenantId || authedToken.tenantId === '') {
+      throw new HttpErrors.BadRequest('BadRequest: tenant id is not provided');
+    }
+
+    context.bind(WafBindingKeys.Request.KeyTenantId).to(authedToken.tenantId);
```
zhaoqin-github: Besides the tenant ID, we also need to know if the token is admin token or not. Does our code already support it?

zongzw: Theoretically, The token(both adminToken and userToken) is only used in auth*-helpers or at the beginning of the request handling. 
During the following process, I think there's no place to refer to it. 

The adminToken retrieved from v3 doe not work yet. Still in negotiating with loopback-connecter-rest owner to get the best solution instead of a workaround:
https://github.com/strongloop/loopback-connector-rest/issues/131

zhaoqin-github: Maybe my terminology is a little outdated... The request token scope might be against one tenant (a cloud user) or the whole system (a cloud admin). A user can only manipulate the resources in his tenant, and the admin can manipulate resources in all tenants. So our API need to know the scope of the request token in order to correctly handle the request.

zongzw: In fact, the function `authRequest` is just doing it: Only authorized user can pass through. 

Currently, we bind `adminToken` to the `application context`, in which it contains resource categories. 
Because the resource user can use today is the very common ones: nova/networks, we currently fetch resource links from the application context's adminToken(assemble tenant info if necessary) for user's access.

If we want to check the scope of each user, we can bind `userToken` to the `request context`, it's easy to do, but a bigger story. 

However, all the binding/getting/resource accessing should be hidden in `services`(identity/network/compute). Should be transparent to upper caller.

zhaoqin-github: Do you mean all of our requests to OpenStack bind with our admin token, instead of user's token?

zongzw: adminToken is bound to the `application`, used for validating user token. 
No token is bound to `request` now, user token is validated at the entrance of request handling.

PR #83, opened 2019-04-02
Many to many service policy

zhaoqin-github:


app/waf/src/controllers/endpointpolicy.controller.ts
```diff
@@ -209,4 +220,67 @@ export class EndpointpolicyController {
...
+    )
+    id: string,
+  ): Promise<CollectionResponse> {
+    let assocs = await this.serviceEndpointpolicyAssociationRepository.find({
+      where: {
+        endpointpolicyId: id,
+      },
+    });
+
+    let serviceIds = assocs.map(({serviceId}) => serviceId);
```
zongzw: Good writing.

PR #88, opened 2019-04-05
Get token from response header of v3/auth/tokens.

zongzw:


app/waf/package.json
```diff
@@ -62,6 +62,7 @@
     "@types/uuid": "^3.4.4",
     "loopback-connector-postgresql": "^3.5.0",
     "loopback-connector-rest": "^3.4.1",
+    "loopback-connector-rest-hdr": "^3.4.4",
```
zhaoqin-github: Are we planning to use this one until loopback-connector-rest supports response header, right?

zongzw: Yeah, their response is slow... so let's maintain our own packages for a while.

PR #89, opened 2019-04-08
Regulate the services for later calls: add subnets/fixed_ips

zongzw:


app/waf/src/services/network.service.ts
```diff
@@ -51,23 +52,27 @@ export class NetworkDriver {
...
-          return this.networkService.v2CreatePort(
-            networkUrl,
-            userToken,
-            portCreationBody,
-          );
+          if (portParams.fixedIp) {
+            body.port.fixed_ips = [{ip_address: portParams.fixedIp}];
+          }
+
+          body.port.name = 'f5-' + portParams.name;
```
zhaoqin-github: Will this port name be "f5-adcId-" ? Or something else?

zongzw: Right. The `portParams.name` should be assembled with adc ID in caller context. May probably `<adcId>-<type>-<name>`.  
adcId is the adc ID in DB. 
type is the type of port 'mgmt' 'ha' 'ext' 'int'. 
name is optional: 

```
  networks: {
    name?: string;                <- name 
    type: 'mgmt' | 'ext' | 'int' | 'ha';   <- type
    networkId: string;
    fixedIp?: string;
    //floatingIp?: string;
    portId?: string; // cannot be appointed.
    ready?: boolean; // cannot be appointed.
    //vips?: [string]; // cannot be appointed.
  }[];
```

PR #90, opened 2019-04-08
Launch api spec design and code frame. 

zongzw:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -193,4 +207,143 @@ export class AdcController {
...
+    ]).then(async ([networkHelper, userToken]) => {
+      for (let k of Object.keys(adc.networks)) {
+        let net = adc.networks[k];
+        if (net.portId && net.ready) continue;
+
+        net.ready = false;
+
+        let portParams: PortCreationParams = {
+          networkId: net.networkId,
+          regionName: 'RegionOne',
```
Niklaus-xie: is regionName a must-have?  seems to me it is not needed.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -193,4 +207,143 @@ export class AdcController {
...
+  private async serialize(adc: Adc, data?: object) {
+    if (data) Object.assign(adc, data);
+    await this.adcRepository.update(adc);
+  }
+
+  private async createOn(adc: Adc): Promise<void> {
+    try {
+      await this.serialize(adc, {status: 'BUILDING'})
+        .then(async () => await this.cNet(adc))
+        .then(async () => await this.cSvr(adc));
```
Niklaus-xie: for example, if cNet() succeeds and cSvr() fails , what happens to the created net-related staff(ports)?  will they just be left un-deleted?

zongzw: To the question: yes, the created ports will be left here undeleted. 

Reason for this: 
1. The recovery logic may be complicated, so I prefer to put it separately somewhere else. 
    In fact, we already support doing that via `ready`, `name` and `status` variables. With those variables, we are able to find any allocated resources. 
2. We make it as possible that the functions(`cNet` & `cSvr`) are reentrant and stateless(save intermediate states back to DB: `this.serialize`). 
    So with the above logic, the user can recall API the second time, then we *continue* the creation without recreating already-allocated resources.

app/waf/src/controllers/adc.controller.ts
zongzw: In the PR: https://github.com/F5Networks/f5-adcaas-openstack/pull/106, `regionName` was hidden. That should answer your(and my) expectation. 

In actual usage, after auth* with keystone, the endpoints are returned within token's `catalog` always tied with a region tag, like: 

```
              {
                region_id: 'RegionOne',
                url:
                  'http://localhost:8774/v2/fde45211da0a44ecbf38cb0b644ab30d',
                region: 'RegionOne',
                interface: 'public',
                id: 'c776745dbad64475a020e219b636894d',
              },
```

So in the above-mentioned PR, I hide it into the auth token handling(with a default value 'RegionOne'). Users can still appoint it if it's not 'RegionOne'.

app/waf/src/controllers/adc.controller.ts
Niklaus-xie: at that moment, those ports are not bound to anything in neutron/nova, so e.g. they can be deleted in neutron (although you saved into our db). when the user calls our api for the 2nd time, they won't be created again according to "ready"/"status" (thus, no related port will be ready in neutron side). this will probably cause consistency issue between our db and neutron/nova.  seems a problem we need to handle anyway in the future.

app/waf/src/controllers/adc.controller.ts
Niklaus-xie: why would user appoint region information during create-port ?

app/waf/src/controllers/adc.controller.ts
zongzw: In the first phase, we suppose users won't manually modify the resources(allocated for & belong to F5 [well named via `portName`]) outside of waf, in which case the resources' status becomes difficult to control. 
However, we still provide the cleanup function to clean all resources. If one port is unable to cleanup, it should be considered to have been removed or re-used by the user themselves.

app/waf/src/controllers/adc.controller.ts
zongzw: The region info is not needed for creating port. but needed to get the network endpoint. 

```
  async createPort(
    userToken: string,
    portParams: PortCreationParams,
  ): Promise<PortResponse> {
    try {
      // TODO: exception check.
      return await this.networkEndpoint(portParams.regionName)
        .then(async networkUrl => {
          let url = networkUrl + '/v2.0/ports';
          let body: PortsRequest = {
            port: {
              network_id: portParams.networkId,
            },
          };
```

It will be hidden soon. What's more, it's not appointed by user. 

See here: 
```
  private async cNet(adc: Adc): Promise<void> {
    await Promise.all([
      this.wafapp.get(WafBindingKeys.KeyNetworkDriver),
      this.reqCxt.get(WafBindingKeys.Request.KeyUserToken),
    ]).then(async ([networkHelper, userToken]) => {
      for (let k of Object.keys(adc.networks)) {
        let net = adc.networks[k];
        if (net.portId && net.ready) continue;

        net.ready = false;

        let portParams: PortCreationParams = {
          networkId: net.networkId,
          regionName: 'RegionOne',                                                 <- not from user. 
          name: <string>(adc.id + '-' + net.type + '-' + k),
        };
```

Niklaus-xie: what if region != "RegionOne" in an  actual environment?

zongzw: As we talked f2f, putting into configuration file/environment is reasonable. 
Prefer to do it in a separate PR.

zongzw:
> The commit history looks a little confused. Can we make it a little more clear?

The commit history is clear now. 
The coverage is OK.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -193,4 +207,143 @@ export class AdcController {
...
+
+          // TODO: Create VM in async way.
+          return await this.createOn(adc).then(async () => {
+            return {id: adc.id};
+          });
+        } catch (error) {
+          throw new HttpErrors.BadRequest(error.message);
+        }
+
+      case 'delete':
```
zhang-shengping: add something in future？

zongzw: Yeah, we will implement delete and setup in the future.

PR #96, opened 2019-04-08
Refactor the action/condition/rule/endpointpolicy

jikui:
Refactor APIs of endpointpolicy/rule/condition/action

### DIFF
--- a/app/waf/src/controllers/endpointpolicy.controller.ts
+++ b/app/waf/src/controllers/endpointpolicy.controller.ts
@@ -23,9 +23,16 @@ import {
   ServiceEndpointpolicyAssociationRepository,
 } from '../repositories';
 import {Schema, Response, CollectionResponse} from '.';
-import uuid = require('uuid');
 
 const prefix = '/adcaas/v1';
+
+const createDesc: string = 'Endpointpolicy resource that need to be created';
+const updateDesc: string =
+  'Endpointpolicy resource properties that need to be updated';
+
+const createRuleDesc: string = 'Rule resource that need to be created';
+const updateRuleDesc: string =
+  'Rule resource properties that need to be updated';
 
 export class EndpointpolicyController {
   constructor(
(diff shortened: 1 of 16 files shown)


app/waf/src/controllers/endpointpolicy.controller.ts
```diff
@@ -23,10 +23,17 @@ import {
   ServiceEndpointpolicyAssociationRepository,
 } from '../repositories';
 import {Schema, Response, CollectionResponse} from '.';
-import uuid = require('uuid');
 
 const prefix = '/adcaas/v1';
 
+const createDesc: string = 'Endpointpolicy resource that need to be created';
```
zongzw: These statements are used only once, declare them while using should be better. Needn't declare them separately.

app/waf/src/models/action.model.ts
```diff
@@ -13,7 +13,12 @@ export class Action extends CommonEntity {
   @property({
     type: 'string',
     required: false,
-    default: 'request',
```
zhaoqin-github: why remove this default value?

app/waf/src/controllers/rule.controller.ts
```diff
@@ -22,9 +22,20 @@ import {
   ConditionRepository,
   ActionRepository,
 } from '../repositories';
-import uuid = require('uuid');
+import {Schema, Response, CollectionResponse} from '.';
 const prefix = '/adcaas/v1';
 
+const createDesc: string = 'Rule resource that need to be created';
```
zongzw: So do these.

app/waf/src/controllers/endpointpolicy.controller.ts
jikui: ok for me.

app/waf/src/models/action.model.ts
```diff
@@ -55,13 +66,23 @@ export class Action extends CommonEntity {
...
+      required: true,
+      example: 'httpUri',
+    },
   })
   type: string;
 
   @property({
     type: 'string',
-    required: true,
+    required: false,
```
zhaoqin-github: It should be required by repository layer, and is not required in API layer, right? We may have two styles of API to create 'has-many' resources. For example, In API layer, POST /policy/{policyId}/rules may not require policyId in body, but POST /rules does require plicyId in body. However, in db layer, plicyId must be required, since it is reference to policy.

app/waf/src/models/endpointpolicy.model.ts
```diff
@@ -1,9 +1,24 @@
-import {model, hasMany} from '@loopback/repository';
+import {model, property, hasMany} from '@loopback/repository';
 import {CommonEntity} from '.';
 import {Rule} from '../models';
 
 @model()
 export class Endpointpolicy extends CommonEntity {
+  @property({
```
zhaoqin-github: Needn't add this column to policy table, since rule table has policy id.

app/waf/test/acceptance/rule.controller.acceptance.ts
```diff
@@ -28,30 +27,31 @@ describe('RuleController', () => {
   });
 
   it('post ' + prefix + '/rules: with id', async () => {
-    const rule = new Rule(
-      createRuleObject({
-        id: uuid(),
-        endpointpolicyId: uuid(),
-      }),
-    );
+    const rule = createRuleObject({
```
zongzw: good.

app/waf/src/models/rule.model.ts
```diff
@@ -1,32 +1,25 @@
...
+import {model, property, hasMany} from '@loopback/repository';
+import {CommonEntity} from '.';
 
 @model()
-export class Rule extends Entity {
+export class Rule extends CommonEntity {
   @property({
     type: 'string',
-    id: true,
+    required: false,
```
zhaoqin-github: It is required.

app/waf/src/controllers/rule.controller.ts
jikui: I will make the change next time.

PR #100, opened 2019-04-09
modify based on the review suggestions

jikui:
modify based on the review suggestions


！
```diff
@@ -0,0 +1,9 @@
+
```
zhaoqin-github: what is it?

jikui: a stale file and  I have deleted it.

PR #103, opened 2019-04-11
Refactor WAF policy API

zhaoqin-github:


app/waf/src/controllers/wafpolicy.controller.ts
```diff
@@ -12,15 +12,18 @@ import {
...
 } from '@loopback/rest';
 import {Wafpolicy} from '../models';
 import {WafpolicyRepository} from '../repositories';
-import uuid = require('uuid');
+import {Schema, Response, CollectionResponse} from '.';
+
 const prefix = '/adcaas/v1';
 
+const createDesc = 'WAF Policy resource that need to be created';
+const updateDesc = 'WAF Policy resource properties that need to be updated';
```
zongzw: Is it possible to in-line the variables, like the L37? 
1. They are used only once. 
2. From the view concern, it should be easier to check when we go to that line.

app/waf/src/models/as3.model.ts
```diff
@@ -249,7 +249,7 @@ export class AS3Action extends AS3Object {
       if (!isNullOrUndefined(this.wafpolicy)) {
         let wafid = <classwafpolicy>this.wafpolicy;
         for (let onewaf of this.wafs) {
-          if (onewaf.id === wafid.wafpolicy) {
+          if (onewaf.id === wafid.wafpolicy && onewaf.name) {
```
zongzw: onewaf -> waf?

app/waf/src/models/as3.model.ts
```diff
@@ -372,7 +372,9 @@ export class AS3WAFPolicy extends AS3Object {
     super();
 
     this.class = 'WAF_Policy';
-    this.name = policy.name;
+    if (policy.name) {
```
zongzw: Just wondering what's the else case..

AS3WAFPolicy's name is declared as `string`, is it possible to be `undefined`?

If it's OK, can we use `this.name = <string>policy.name;` instead of an `if`?

app/waf/src/models/as3.model.ts
zhaoqin-github: It will be removed by our new as3 translation code in next commit. Needn't to improve it any more.

app/waf/src/models/as3.model.ts
zhaoqin-github: It will be removed by our new as3 translation code in next commit. Needn't to improve it any more.

app/waf/src/controllers/wafpolicy.controller.ts
zhaoqin-github: Done

PR #104, opened 2019-04-11
AS3 translation

zhaoqin-github:


app/waf/src/controllers/application.controller.ts
```diff
@@ -175,9 +197,115 @@ export class ApplicationController {
...
+    }
+  }
+
+  private async loadAction(action: Action): Promise<void> {
+    if (action.type === 'waf' && action.policy) {
+      action.wafpolicy = await this.wafpolicyRepository.findById(action.policy);
+    }
+  }
+
   @post(prefix + '/applications/{applicationId}/declarations', {
```
zongzw: To my thinking, Declaration should have its own controller, instead of applications. 
Also: `'/applications/{applicationId}/declarations`: I think we needn't an application id to index declaration. 

The relation of declaration and application is 'related' not 'belong to'. 

Users can deploy the declaration, which already has no relation to the application. 

Think that if the user changes the application-tree again, and deploy an older declaration with API: '/applications/{applicationId}/declarations' will make users confused, application-tree is newer than declaration I am deploying.

app/waf/src/models/application.model.ts
```diff
@@ -12,13 +12,82 @@ export class Application extends CommonEntity {
...
+  @property({
+    type: 'string',
+    required: false,
+    schema: {
+      update: true,
+      response: true,
+      example: '11111111-2222-3333-4444-555555555555',
+    },
+  })
+  currentDeclarationId?: string;
```
zongzw: It will produce data redundancy. you have to handle the setting/maintain on the value for application changes or new declaration.
In fact, this information can be gotten via querying declaration.

app/waf/src/models/application.model.ts
```diff
@@ -12,13 +12,82 @@ export class Application extends CommonEntity {
...
+    type: 'string',
+    required: false,
+    schema: {
+      create: true,
+      update: true,
+      response: true,
+      example: '11111111-2222-3333-4444-555555555555',
+    },
+  })
+  adcId?: string;
```
zongzw: Adc and application can be decoupled. I think there's no need to make an adcId in application. 

Suppose users deploy their configuration by running `deploy(adcId, declarationId)`, it's simple.

app/waf/src/controllers/application.controller.ts
```diff
@@ -299,96 +437,50 @@ export class ApplicationController {
     responses: {
       '204': Schema.emptyResponse('Successfully deploy Application resource'),
       '404': Schema.notFound('Can not find Application resource'),
+      '422': Schema.unprocessableEntity('Fail to deploy Application resource'),
     },
   })
   async deployById(
```
zongzw: Move to the declaration.controller is better since what users deploy is a declaration, not an application.

app/waf/src/models/as3.model.ts
```diff
@@ -1,387 +1,47 @@
...
   declaration: AS3Declaration;
 
-  constructor(params: {[key: string]: Object}) {
-    super();
-
-    let adc = <Adc>params['adc'];
-
-    this.class = 'AS3';
-    this.action = 'deploy';
+  constructor(adc: Adc, application: Application, declaration?: Declaration) {
```
zongzw: Again, I think `constructor(adc: Adc, declaration?: Declaration) {` is better.

app/waf/src/models/as3.model.ts
```diff
@@ -1,387 +1,47 @@
...
+    if (adc.name) {
+      this.declaration.label = adc.name;
     }
 
-    let members = <Member[]>params.members;
-
-    this.members = [];
-    for (let member of members) {
-      this.members.push(new AS3PoolMember(member));
+    if (adc.description) {
```
zongzw: `if` can be removed: 
if `adc.description` is not null, this.declaration.remark is adc.description. 
if `adc.description` is null, this.declaration.remark would be null. 
No difference. Same to adc.name.

app/waf/src/models/as3.model.ts
```diff
@@ -1,387 +1,47 @@
...
-  name: string;
-  url: string;
-
-  constructor(policy: Wafpolicy) {
-    super();
-
-    this.class = 'WAF_Policy';
-    if (policy.name) {
-      this.name = policy.name;
+    if (declaration) {
```
zongzw: if no declaration(declaration is undefined), throw exception directly. 
We don't allow deployment with no declaration go forwards.

app/waf/src/controllers/application.controller.ts
```diff
@@ -299,96 +437,50 @@ export class ApplicationController {
...
+    let req = new AS3DeployRequest(adc, application, declaration);
+    return await this.as3Service.deploy(AS3_HOST, AS3_PORT, req);
+  }
 
-    let adcs = await this.adcRepository.find({
-      where: {
-        id: tenantAssocs[0].adcId,
-      },
-    });
+  @post(prefix + '/applications/{applicationId}/cleanup', {
```
zongzw: Move to declaration.controller. 

Or it makes confusion:

@post(prefix + '/applications/{applicationId}/cleanup'
is too easy to be thought of 
@del(prefix + '/aplications/{applicationId}').

app/waf/src/controllers/application.controller.ts
```diff
@@ -299,96 +437,50 @@ export class ApplicationController {
...
+    responses: {
+      '204': Schema.emptyResponse('Successfully cleanup Application resource'),
+      '404': Schema.notFound('Can not find Application resource'),
+    },
+  })
+  async cleanupById(
```
zongzw: As to the API spec, I prefer
```
<obj>/{objId}/action:
body: {
    <actionName>: {parameters}
}
```

As we talked before, following OpenStack endpoint standard.

app/waf/test/acceptance/application.controller.acceptance.ts
```diff
@@ -144,27 +133,30 @@ describe('ApplicationController', () => {
...
+      let application = await givenApplicationData(wafapp, {
+        adcId: adc.id,
+      });
+
+      await client
+        .post(prefix + '/applications/' + application.id + '/declarations')
+        .send({name: 'a-declaration'})
+        .expect(200);
+
+      deployStub.returns(Promise.resolve('Hello'));
```
zongzw: The deloyStub makes the acceptance not as pure as we expected. 

Along time, there's more and more API call to AS3, I prefer to start up a rest server to mock the response outside, just like mock-<openstack> does. 
It's up to you.

app/waf/src/models/common.model.ts
```diff
@@ -1,6 +1,15 @@
 import {Entity, property} from '@loopback/repository';
+import {MetadataInspector} from '@loopback/metadata';
+import {MODEL_PROPERTIES_KEY, PropertyDefinition} from '@loopback/repository';
+import {AS3Declaration} from '.';
+
+function as3Name(postfix: string) {
```
zongzw: It changes the user's given names and unrecoverable. 
We may need to find a way to solve it, etc. encoding with base64 or so..

app/waf/src/models/declaration.model.ts
```diff
@@ -21,11 +21,9 @@ export class Declaration extends CommonEntity {
       example: {},
     },
   })
-  content: AS3JSONObject;
+  content: object;
```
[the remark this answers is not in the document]
zhaoqin-github: Let's implement that in our next pr.

app/waf/src/controllers/application.controller.ts
```diff
@@ -187,13 +315,23 @@ export class ApplicationController {
     reqBody: Partial<Declaration>,
```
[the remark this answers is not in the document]
zhaoqin-github: Declaration will at least have a name which can be specified by user.

app/waf/src/controllers/application.controller.ts
zhaoqin-github: One declaration only belongs to one application. It can not be association with several applications.

app/waf/src/models/application.model.ts
zhaoqin-github: It is not redundant data. If someone hopes to deploy the config of one application without inputting a specific declaration id, this property is required. Otherwise, everyone need to know exactly which declaration is the right one and which is not.

We can let our app not to automatically specify this id. Just let the user to specify it by himself.

app/waf/src/models/as3.model.ts
zhaoqin-github: I do not hope to add 'label' and 'remark' key into AS3 request body, so that I add the 'if'.

app/waf/src/models/as3.model.ts
zhaoqin-github: A request without declaration is the request to remove the tenant in BIG-IP. That is the behavior we need.

app/waf/src/controllers/application.controller.ts
zhaoqin-github: Maybe we can give a better name to this action.

app/waf/src/models/common.model.ts
zhaoqin-github: This property is the index key of AS3 object, not the object name given by user. But I think change that index key to something like base64 encoded string seems to be a good idea.

zhaoqin-github: AS3 name does not accept '=', we need another way to encode..

app/waf/src/models/as3.model.ts
zhaoqin-github: Remove these lines, since they are useless in fact.

app/waf/src/models/common.model.ts
zhaoqin-github: I change the AS3 name back to 'F5_XXX'. And input resource id to AS3 object label property. Let us look for a better encoding method later.

app/waf/src/controllers/application.controller.ts
zongzw: I didn't say that the declaration can be associated with multiple application. 

"One declaration only belongs to one application." I may not agree with this statement. 

it's not a belong-to relation, but a related. In fact, the declaration is related with not only application but kinds of other configurations. 

The declaration should have its own controller/model, also endpoint prefix. I think. That is my core thinking/main concern.

app/waf/src/controllers/application.controller.ts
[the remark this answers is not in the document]
zongzw: There should be a way to tell user that what's needed, what's not. 

Here, 

name is required. 
content MUST NOT be provided.

app/waf/src/controllers/application.controller.ts
zongzw: I prefer to move declarations REST APIs to /declarations (declaration.controller.ts).

app/waf/src/models/application.model.ts
zongzw: I think requiring the user to generate the declaration before deploying it will make the workflow clear. 
That means `deploy` action belongs to declaration. Not application.

app/waf/src/models/declaration.model.ts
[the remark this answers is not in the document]
zongzw: If so, please add a TODO.

app/waf/src/controllers/application.controller.ts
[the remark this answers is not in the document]
zhaoqin-github: Those has been done by model property decorator. API validation has not been implemented. That is the scope of other stories.

app/waf/src/controllers/application.controller.ts
zhaoqin-github: In our resource model, this 'declaration' terminology just mean the declaration of an application. We will not create or deploy any other kind of AS3 declaration.

app/waf/src/models/application.model.ts
zhaoqin-github: Both of them should have deploy action. In this change set, we only implement one of them.

app/waf/src/controllers/application.controller.ts
zhaoqin-github: Done

app/waf/src/controllers/application.controller.ts
zongzw: Got it.

app/waf/src/controllers/application.controller.ts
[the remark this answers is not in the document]
zongzw: OK.

app/waf/src/models/application.model.ts
zongzw: Still, feel that letting `application` to have only the configuration API set and move deploying operations to `declaration` makes the frame simpler. 
Up to you. Got your meaning.

app/waf/src/models/declaration.model.ts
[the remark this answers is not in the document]
zhaoqin-github: Done

app/waf/src/controllers/application.controller.ts
```diff
@@ -175,9 +197,115 @@ export class ApplicationController {
     await this.applicationRepository.deleteById(id);
   }
 
+  private async loadApplication(app: Application): Promise<void> {
```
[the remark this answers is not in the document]
zhaoqin-github: Resolve this thread, since it is obsolete.

app/waf/src/controllers/application.controller.ts
zhaoqin-github: We have not yet made a decision about resource action API style. OpenStack resource action API is not very good, because it will have us to create some extra coding (eg. type definition), just like what you did for ADC resource. Personally, I will prefer a generic way, for example:

```
POST /resource-url-path/action
{
  "action": "deploy",
  "paramaters": {
    "a": 1,
    "b": 2
  }
}
```

The action response can be 204, resource content, or a specific response body. In this way, we will not need to define every action type in code. What do you think?

zhaoqin-github: And another alternative is to declare the action name in path, as what I did in this PR.

```
POST /resource-url-path/{action_name}
{body is optional}
```

Its advantage is that it will be flexible for us to define the request/response schema for every action APIs. Our previous two options both register action handers with one fixed url path. That means we have to define the same request/response schema for all action APIs. It will be a little hard to generate very detail action API document.

app/waf/src/models/application.model.ts
zhaoqin-github: The deploy API need to support an adcId parameter, in order to deploy the configuration to a user specified device. It also need a default device id, so that user can deploy the config without inputing a device id.

app/waf/src/models/as3.model.ts
zhaoqin-github: Some information of AS3 request if from Application resource, for example, the application name.

PR #109, opened 2019-04-18
Trust ADC hardware

zhaoqin-github:


app/waf/src/services/trusteddevice.service.ts
```diff
@@ -0,0 +1,53 @@
...
+  trust(host: string, port: number, body: object): Promise<TrustedDevices>;
+  query(host: string, port: number, deviceId: string): Promise<TrustedDevices>;
+  untrust(
+    host: string,
+    port: number,
+    deviceId: string,
+  ): Promise<TrustedDevices>;
+}
+
+export class TrustedDeviceServiceProvider
```
zhang-shengping: Does any place use this Provider?

zhaoqin-github: It will be called by LB4 framework.

PR #111, opened 2019-04-18
Install AS3 RPM on ADC device

zhaoqin-github:


app/waf/test/acceptance/adc.controller.acceptance.ts
```diff
@@ -423,9 +446,281 @@ describe('AdcController', () => {
...
+      let response = await client
+        .post(prefix + '/adcs')
+        .set('X-Auth-Token', ExpectedData.userToken)
+        .set('tenant-id', ExpectedData.tenantId)
+        .send(adc)
+        .expect(200);
+
+      expect(response.body.adc).to.containDeep(toJSON(adc));
+
+      await sleep(50);
```
zongzw: Left it here, I'll change it away.

zhaoqin-github: After checkAndWait() can support error path, we will not need to sleep here.

PR #117, opened 2019-04-27
Add onboarding logic and tests.

zongzw:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -250,16 +255,27 @@ export class AdcController {
...
-          // DO here.
-          return {id: adc.id};
+          try {
+            this.logger.debug('start to do onbarding');
+            let doMgr = await OnboardingManager.instanlize();
+            let doBody = await doMgr.assembleDo(adc);
+            this.logger.debug(
+              'Json used for onboarding: ' + JSON.stringify(doBody),
+            );
+            let doRes = await doMgr.onboarding(doBody);
```
zhaoqin-github: How long will it take? Will we plan not to wait here in the future?

app/waf/src/models/adc.model.ts
```diff
@@ -148,11 +176,86 @@ export class Adc extends CommonEntity {
...
 
-  // @property({
-  //   type: 'object',
-  //   required: true,
-  // })
-  // onBoarding: ConfigBody['onBoarding'];
+  @property({
+    type: 'object',
+    schema: {
+      create: true,
```
zhaoqin-github: I do not think API can allow to input this property.

app/waf/src/models/adc.model.ts
```diff
@@ -148,11 +176,86 @@ export class Adc extends CommonEntity {
...
+      response: true,
+      example: {
+        declaration: {
+          system: {
+            license: {
+              regKey: 'Q2431-06294-92275-11650-8028186',
```
zhaoqin-github: It is OK to let user to retrieve onboarding data. That would be a useful way for them to quickly know the configuration of their device. However, we may need to consider to hide some information. For example, I am not quite if we can let a tenant user know the license key. Maybe not.

zhang-shengping:
It looks good to me, I can not find any simple bugs or explicit error.

app/waf/src/controllers/adc.controller.ts
zongzw: Generally, it can be over in < 20s. I added a 'TODO' in the code to think it in the future whether to async execute it.

app/waf/src/models/adc.model.ts
zongzw: Done. removed it as we discussed in the sync up meeting.

app/waf/src/models/adc.model.ts
zongzw: Yeah, we can support retrieving some of the information. However, after all we need not user to provide the detailed information.

app/waf/.vscode/launch.json
```diff
@@ -21,7 +21,12 @@
       "runtimeExecutable": null,
       "internalConsoleOptions": "openOnSessionStart",
       "env": {
-        "NODE_ENV": "production"
+        "NODE_ENV": "production",
```
zongzw: Will remove `.vscode` later.

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: OK

app/waf/src/models/adc.model.ts
zhaoqin-github: Thanks

zhang-shengping:
I add a tenantId to this file `app/waf/test/fixtures/controllers/mocks/mock.openstack.controller.ts`, you may need to add it as well.

PR #120, opened 2019-05-07
Add utils.ts and waitAndCheck function.

zongzw:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -458,12 +476,15 @@ export class AdcController {
       this.reqCxt.get(WafBindingKeys.Request.KeyUserToken),
       this.reqCxt.get(WafBindingKeys.Request.KeyTenantId),
     ]).then(async ([computeHelper, userToken, tenantId]) => {
-      let rootPass = Math.random()
-        .toString(36)
-        .slice(-8);
-      let adminPass = Math.random()
-        .toString(36)
-        .slice(-8);
+      // TODO: uncomment me.
```
zongzw: Temporarily comment on the password generation since we have lots of manual tests which needs the password inputting, restore them to 'default' and 'admin' would be easy for current developments.

app/waf/src/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+  if (tryTimes <= 0) {
+    utilsLogger.debug(`'${funcName}' timeout.`);
+    return Promise.reject();
+  }
+
+  let hdlr = async (b: boolean): Promise<boolean> => {
+    if (typeof b === 'boolean' && b) {
+      utilsLogger.debug(`'${funcName}' returns ${b}, countdown: ${tryTimes}`);
+      return Promise.resolve(true);
+    } else throw new Error();
```
zhang-shengping: Do any comments need in this Error?

app/waf/src/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  try {
+    await checkFunc(...funcArgs).then(hdlr, hdlr);
```
zhang-shengping: I do not know is there any boolean parameters will be passed to this async function?

app/waf/package.json
```diff
@@ -66,7 +66,8 @@
     "plural": "^1.1.0",
     "typescript-logging": "^0.6.3",
     "uuid": "^3.3.2",
-    "ip": "^1.1.5"
+    "ip": "^1.1.5",
+    "@network-utils/tcp-ping": "^1.1.0"
```
zhang-shengping: are these package safe?

app/waf/src/utils.ts
zongzw: Agree, it needs some information output indeed, but it may be not suitable to transfer the information upwards, it's a loop function until timeout. So I make it output to the logger. 

Thus the log seems like: 

```
2019-05-09 08:06:19,012 DEBUG [utils.logger] Check and wait 'bigipReady' to response, countdown: 206.
2019-05-09 08:06:19,018 DEBUG [utils.logger] 'bigipReady' response with Error: Host unreachable: {"ipaddr":"10.250.15.120","port":443}, countdown: 206
```

app/waf/src/utils.ts
zongzw: Didn't get your point...
The `checkFunc` is supposed to return `Promise<boolean>`, which is resolved/rejected to `hdlr`, then return nothing in `checkAndWait` context. 
And the `checkAndWait` function returns `Promise<void>`(resolve when `checkFunc` condition is fulfilled; reject when `timeout`.)

app/waf/package.json
zongzw: They are both safe. But you remind me a lot. 

After the scan with `snyk`, we found the following vulnerabilities. 
`snyk` scan may need to be added to our CI process. It support github.io integration.
@zhaoqin-github 

```
Analyzing npm dependencies for /Users/zong/GitRepos/zongzw/f5-adcaas-openstack/app/waf
Querying vulnerabilities database...
Tested 248 dependencies for known vulnerabilities, found 1 vulnerabilities, 2 vulnerable paths.

? ✗ High severity vuln found in js-yaml@3.13.0, introduced via @loopback/rest@1.9.1
    Description: Arbitrary Code Execution
    Info: https://snyk.io/vuln/SNYK-JS-JSYAML-174129
    From: @loopback/rest@1.9.1 > js-yaml@3.13.0

  Remediation options Skip

? ✗ High severity vuln found in js-yaml@3.13.0, introduced via @loopback/rest-explorer@1.1.13
    Description: Arbitrary Code Execution
    Info: https://snyk.io/vuln/SNYK-JS-JSYAML-174129
    From: @loopback/rest-explorer@1.1.13 > @loopback/rest@1.9.1 > js-yaml@3.13.0

  Remediation options Skip
```

PR #123, opened 2019-05-09
Add ntp onboarding.

zongzw:


app/waf/src/services/do.service.ts
```diff
@@ -155,6 +155,25 @@ export class OnboardingManager {
...
+        let ntpData = {
+          class: 'NTP',
+          servers: ['0.pool.ntp.org', '1.pool.ntp.org', '2.pool.ntp.org'],
+          timezone: 'UTC',
+        };
+        this.logger.debug('Add new ntp.');
+        return Object.assign(target, {myNtp: ntpData});
+      } catch (error) {
+        this.logger.debug('No ntp found.');
+        return target;
```
zhang-shengping: does not need to throw, right?

zongzw: Yeah, make it behaves like the onboarding body assembling on ntp did not happen.

zongzw: But should make the error.message print into debug logs. 
Will do it in another PR.

PR #124, opened 2019-05-09
add mutltitenancy to service

zhang-shengping:
add mutltitenancy to service

### DIFF
--- a/app/waf/test/acceptance/service.controller.acceptance.ts
+++ b/app/waf/test/acceptance/service.controller.acceptance.ts
@@ -1,44 +1,87 @@
 import {Client, expect, toJSON} from '@loopback/testlab';
 import {WafApplication} from '../..';
-import {setupApplication, teardownApplication} from '../helpers/test-helper';
+import {
+  setupApplication,
+  teardownApplication,
+  TestingApplication,
+  setupRestAppAndClient,
+  RestApplicationPort,
+  teardownRestAppAndClient,
+} from '../helpers/test-helper';
 import {
   givenEmptyDatabase,
   givenServiceData,
 } from '../helpers/database.helpers';
+import {
+  ShouldResponseWith,
+  MockKeyStoneController,
+  ExpectedData,
+} from '../fixtures/controllers/mocks/mock.openstack.controller';
 import uuid = require('uuid');
 
 describe('ServiceController', () => {
   let wafapp: WafApplication;
   let client: Client;
+  let mockKeystoneApp: TestingApplication;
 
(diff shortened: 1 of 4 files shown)


app/waf/test/acceptance/service.controller.acceptance.ts
```diff
@@ -1,44 +1,87 @@
...
 import uuid = require('uuid');
 
 describe('ServiceController', () => {
   let wafapp: WafApplication;
   let client: Client;
+  let mockKeystoneApp: TestingApplication;
 
   const prefix = '/adcaas/v1';
 
+  let envs: {[key: string]: string} = {
```
zongzw: I'm thinking that when there are multiple usages on this, could we extract it out to `test-helper.ts`.

app/waf/test/acceptance/service.controller.acceptance.ts
```diff
@@ -1,44 +1,87 @@
...
   });
+
   beforeEach('Empty database', async () => {
     await givenEmptyDatabase(wafapp);
   });
 
-  after(async () => {
-    await teardownApplication(wafapp);
-  });
+  let setupEnvs = async () => {
```
zongzw: -> `test-helper.ts`?

app/waf/test/acceptance/service.controller.acceptance.ts
```diff
@@ -1,44 +1,87 @@
...
+    process.env.PRODUCT_RELEASE = '1';
+    for (let env of Object.keys(envs)) {
+      process.env[env] = envs[env];
+    }
+  };
 
-  it('get' + prefix + '/services/count', async () => {
-    const service = await givenServiceData(wafapp, uuid(), {id: uuid()});
+  let teardownEnvs = async () => {
```
zongzw: -> `test-helper.ts`?

app/waf/test/acceptance/service.controller.acceptance.ts
zhang-shengping: Yes, it should be in test-helper.ts, this will affect other tests. Do you think it's better to do it after mutli-tenancy applied to all APIs?

zongzw: OK for that. 
It doesn't affect others, I think. 
Only moving the definition to test-helper.ts, while the call to them is still in *.controller.acceptance.ts.

zhang-shengping: I got your idea. You mean to move the definition to the test-helper.ts. I can change all of these in the next patch.

PR #127, opened 2019-05-13
Do create and setup bigip VE asynchronously.

zongzw:


zongzw:
> hope spaces do not matter.

It's already. npm run prettier:fix helped fixed it.

PR #130, opened 2019-05-13
Add license check control.

zongzw:


app/waf/src/services/bigip.service.ts
```diff
@@ -97,6 +97,26 @@ export class BigIpManager {
...
+    await this.mustBeReachable();
+
+    let url = `${this.baseUrl}/mgmt/tm/sys/license`;
+
+    let response = await this.bigipService.getInfo(url, this.cred64Encoded);
+    let resObj = JSON.parse(JSON.stringify(response))['body'][0];
+    this.logger.debug(`get ${url} resposes: ${JSON.stringify(resObj)}`);
+
+    for (let entry of Object.keys(resObj.entries)) {
+      if (!entry.endsWith('/mgmt/tm/sys/license/0')) continue;
```
zhang-shengping: This is not bad, it sounds useful in the future :D

zongzw: Yeah. However, there may still be some adjustment in the future to adopt the coming uncertainties.

PR #136, opened 2019-05-16
Add more precise control on checking create/setup done.

zongzw:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -380,10 +381,17 @@ export class AdcController extends BaseController {
...
+        `Adc status is ' ${
+          adc.status
+        }. Cannot be operated on, please wait for its finish.`,
+      );
+
     switch (Object.keys(actionBody)[0]) {
       case 'create':
         if (adc.status !== 'NONE' && adc.status !== 'POWERERR')
-          throw new HttpErrors.BadRequest(
+          throw new HttpErrors.UnprocessableEntity(
```
zhaoqin-github: When I looking at this, I think about resource action api again. ```POST /{id}/{action_name}``` seems to be better than ```POST /{id}/action```, because the 1st one is flexible for us to define action API schema. The 2nd one accommodates all action API in one schema, that schema may be huge. If some action API has difference request/response schema, that API reference might be a little confusing.

zongzw: Since we are integrating with OS, I think it's better to follow the same style of the API definition. That's our agreement on this point. 

Selecting `POST /{id}/{action_name}` just because it's easier to define schema. 

Furthermore, currently, the ADC CRUD API is simple, which the schema would be huge.

zongzw:
> Need a little more time to look into the code. Generally, this change set is attempting to do some checking and raise error, so we may need corresponding test to ensure those checking and error handling works. Otherwise, we are not able to prevent other change sets breaking those checks.

I agree that we do the test on exception. Could I do it separately, because: 
I have 3 PRs, after they are combined(especially ADC state controller), the exception handlings can be unified and well managed.

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: I mean it may not be flexible to define action api schema, if we combine several action api into one url path. Another benefit might be that it will be a little easy to separate action api handler code. I am not pursuing to change these code right now. Let's think it again later.

PR #137, opened 2019-05-16
Add adc state controller and usage.

zongzw:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -588,3 +580,132 @@ type AddonReqValues = {
...
+  }
+}
+
+type AdcStateEntry = {
+  state: string;
+  check: Promise<boolean>;
+  next: string[];
+};
+
+enum AdcState {
```
zhaoqin-github: Should we add a new state, eg. 'ACTIVE'? After  TRUSTED, we can install AS3, then ADC can move to ACTIVE state, which indicates this device is ready to process AS3 request.

zhaoqin-github: I am not sure if 'NEW' is better than 'NONE' to indicate that state. Maybe 'NONE' make someone a little confused...

app/waf/src/controllers/adc.controller.ts
```diff
@@ -588,3 +580,132 @@ type AddonReqValues = {
   userToken: string;
   tenantId: string;
 };
+
+class AdcStateCtrlr {
+  private defaultCheck = (): Promise<boolean> => {
+    return Promise.resolve(true);
+  };
+  private states: AdcStateEntry[] = [
```
zhaoqin-github: Can we draw an ascii diagram about the state transition.....

app/waf/src/controllers/adc.controller.ts
zongzw: Yes, we should have a final state `ACTIVE` after all the works are done. 
In the short term, we may have extension installation after trust. 
I suggest to sort out the states incrementally. That's to add such works in later PRs.

app/waf/src/controllers/adc.controller.ts
zongzw: Yes, we need that. 
But it should be delay some time since doing it right now would burden us on the consistency between code and ASCII docs.

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: 3ks

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: It will require some effort to update doc when code changes. Nothing can help to accelerate that. Just hope to say a design spec might be required. A ascii doc might be better than a ppt, since it is easy for other persons to maintain and contribute.

zhaoqin-github: Maybe we can have a small review meeting for this pr, so that you can let other persons to know your design concept.

zongzw: Can I draft it as another PR, then reviewed with all f2f to populate this concept.?

app/waf/src/controllers/adc.controller.ts
```diff
@@ -334,47 +335,43 @@ export class AdcController extends BaseController {
...
-      port: adc.management!.tcpPort,
-    });
-    let bigipReady = async (): Promise<boolean> => {
-      return await bigipMgr.getSys().then(() => {
-        return true;
-      });
-    };
-    await checkAndWait(bigipReady, 240).then(
+    await checkAndWait(
+      () => this.adcStCtr.readyTo(AdcState.ONBOARDED).then(() => true),
```
Niklaus-xie: is this line and line 352 duplicated.

zongzw: Right. Done for that. Thanks.
The usage of StateController is not made fully. There are 2 places which are using it. Later will apply all over the code.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -641,3 +629,133 @@ type AddonReqValues = {
...
 };
+
+class AdcStateCtrlr {
+  private defaultCheck = (): Promise<boolean> => {
+    return Promise.resolve(true);
+  };
+  private states: AdcStateEntry[] = [
+    {
+      state: AdcState.NEW,
+      check: Promise.resolve(true),
```
Niklaus-xie: did you intend to use this.defaultCheck() here?

app/waf/src/controllers/adc.controller.ts
Niklaus-xie: 👌

app/waf/src/controllers/adc.controller.ts
zongzw: Planned to, but in fact, no use case found. 
Done to remove `defaultCheck`, will add it back if necessary.

Niklaus-xie: 👌， removal seems fine.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -641,3 +629,130 @@ type AddonReqValues = {
...
+  check: Promise<boolean>;
+  next: string[];
+};
+
+enum AdcState {
+  NEW = 'NEW',
+
+  POWERON = 'POWERON',
+  POWERING = 'POWERING',
+  POWERERR = 'POWERERR',
```
zhaoqin-github: Will you change ERR to ERROR?

zongzw: Separate PR.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -641,3 +629,130 @@ type AddonReqValues = {
...
+    return true;
+  }
+
+  private async trusted(): Promise<boolean> {
+    return Promise.resolve(true);
+  }
+
+  private async onboarded(): Promise<boolean> {
+    let bigipMgr = await this.getBigipMgr();
+    return Promise.all([
```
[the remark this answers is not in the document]
zongzw: Maybe, but in reality, the tcp connect is very short. 
Even we plan to provision multiple bigip VEs, it's basically very hard to encounter the situation of all requests happen together in a short timescope. 

But you are right, this should be a concern we may consider via doing some measuring test, how many connections during the whole process and when they happens..

zongzw:
> There are at least 6 new paths are not covered by your testing. Suggest to improve a little.

I may have no time working on this. Will do it separately in another PR in different sprint.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -751,3 +715,131 @@ type AddonReqValues = {
...
+    if (this.adc.management) return false;
+    if (this.adc.compute.vmId) return false;
+    for (let net of Object.keys(this.adc.networks)) {
+      if (this.adc.networks[net].portId) return false;
+    }
+    return true;
+  }
+
+  private async trusted(): Promise<boolean> {
+    return Promise.resolve(true);
```
Niklaus-xie: don't we need to add logic to check if it is trusted or not here?

zongzw: We need to add this part. Because they are in 2 separate PRs, we may need to implement it later.

zhaoqin-github:
Not sure if this code change will impact the demo. You may decide a right time to merge the code.

zongzw:
> Not sure if this code change will impact the demo. You may decide a right time to merge the code.

The demo code has been frozen to docker images. It's OK to merge.

PR #139, opened 2019-05-17
Implement the adc deletion.

zongzw:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -594,6 +595,88 @@ export class AdcController extends BaseController {
...
+        let computeMgr = await this.wafapp.get(
+          WafBindingKeys.KeyComputeManager,
+        );
+        if (adc.compute.vmId) {
+          await computeMgr
+            .deleteServer(addon.userToken, adc.compute.vmId!, addon.tenantId)
+            .then(() => {
+              this.logger.debug(`Deleted the vm ${adc.compute.vmId!}`);
+              delete adc.compute.vmId;
+              adc.management = undefined;
```
zhaoqin-github: Will '{}' be better than 'undefined'?

app/waf/src/models/adc.model.ts
```diff
@@ -125,7 +125,7 @@ export class Adc extends CommonEntity {
       response: true,
     },
   })
-  management: ConfigTypes['management'];
+  management?: ConfigTypes['management'];
```
zhaoqin-github: Is that posible to remove '?' ?

app/waf/src/controllers/adc.controller.ts
zongzw: The properties of adc.management(ipaddr, port, user, pswd..) is not optional, which means, 
adc.management should have 2 types of values: undefined or not undefined(with properties in it.).

app/waf/src/models/adc.model.ts
zongzw: This relates to the upper comments. 
'?' means it can be null at the beginning of the record creation.

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: I did a test just now. Adc.management can be '{}', or just have one or two properties. No problem occurs in the app.

zongzw: When I remove one property 'username', it tells me: 
```[ts] Property 'username' is missing in type '{ password: string; rootPass: string; tcpPort: number; ipAddress: string; }' but required in type '{ ipAddress: string; tcpPort: number; username: string; password: string; rootPass: string; }'.```

How did you test it?

Anyhow, why do you think {} is better than undefined?

I think that undefined means not set, while distinguishing {} and {...}, you have still check its properties..

zhaoqin-github: I use master branch code to create one adc object, for example, posting a '{}' in management property, and I can get that object back from db, whose management property is still '{}'.

zhaoqin-github: Maybe that should be a type checking and will only occur during compiling. The js code will never care that.

zhaoqin-github:
@zongzw do you have plan to test your exception path?

app/waf/src/controllers/adc.controller.ts
zongzw: The code we work on is typescript, which has strict type check. 
After it's compiled to nodejs, it becomes loose on type check. 
That's a core difference between ts and js. 

Anyhow, that's a little out of the topic mentioned here. I'd like to left it undefined here.

PR #140, opened 2019-05-20
Enable multi-tenancy support for Application resource

zhaoqin-github:


app/waf/src/controllers/base.controller.ts
```diff
@@ -9,6 +9,11 @@ export class BaseController {
     @inject(RestBindings.Http.CONTEXT)
     protected reqCxt: RequestContext,
   ) {
+    //TODO: This is a workaround to run UT. Some UT codes call this
```
zongzw: Could you please describe it a little more, I didn't catch this case.
For that supposed case, will we throw an error instead of return?

Generally speaking, the source code should not bend to adopt the UT for any reason. 
Just my initial thinking on this.

zhaoqin-github: In before(), we may call this

```
    controller = await wafapp.get<ApplicationController>(
      'controllers.ApplicationController',
    );
```

and the call stack will look like this

```
  1) ApplicationController
       "before all" hook: setupApplication for "post /adcaas/v1/applications: with no id":
     TypeError: Cannot read property 'get' of undefined
      at new BaseController (src/controllers/base.controller.ts:12:33)
      at new ApplicationController (src/controllers/application.controller.ts:49:5)
      at value_promise_1.transformValueOrPromise.args (node_modules/@loopback/context/src/resolver.ts:73:14)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)
```

PR #149, opened 2019-05-23
Deploy declaration resource

zhaoqin-github:


app/waf/src/controllers/declaration.controller.ts
```diff
@@ -333,4 +342,56 @@ export class DeclarationController extends BaseController {
       .declarations(applicationId)
       .delete({and: [{id: declarationId}, {tenantId: await this.tenantId}]});
   }
+
+  @post(
+    prefix +
+      '/applications/{applicationId}/declarations/{declarationId}/deploy',
```
zongzw: As I recommended before, I don't think to prefix `/applications/{applicationId}` to the API is a good idea. 

Indeed, the declaration has a relation with the application... In fact, the declaration has a relationship with all elements of the TREE... 

The relation link has already been saved in declaration model, we/user can get it through the query string(`GET /declarations?applicationId=xxx` or `GET /applications/{applicationId}/delcarations`). 
That's not meaning we need to bind the relationship everywhere, even into the API call. 

Think that: when the application TREE's configuration is newer than a declaration generated before, to prefix the applicationId to that declaration's deployment would be confused, no matter succeed or fail. 

I think declaration is a standalone existence, which can exist or be operated on with no regarding the application which it is generated from...

app/waf/test/acceptance/declaration.controller.acceptance.ts
```diff
@@ -385,6 +386,115 @@ describe('ApplicationController declaration test', () => {
...
+      prefix +
+      '/applications/{applicationId}/declarations/{declarationId}/deploy: deploy declaration',
+    async () => {
+      const adc = await givenAdcData(wafapp);
+      const application = await givenApplicationData(wafapp, {
+        adcId: adc.id,
+      });
+      const declaration = await givenDeclarationData(wafapp, {
+        applicationId: application.id,
+        tenantId: ExpectedData.tenantId,
```
zhang-shengping: you may delete this `tenantId: ExpectedData.tenantId,`

app/waf/test/acceptance/declaration.controller.acceptance.ts
```diff
@@ -385,6 +386,115 @@ describe('ApplicationController declaration test', () => {
...
+
+  it(
+    'post' +
+      prefix +
+      '/applications/{applicationId}/declarations/{declarationId}/deploy: deploy without default ADC',
+    async () => {
+      const application = await givenApplicationData(wafapp);
+      const declaration = await givenDeclarationData(wafapp, {
+        applicationId: application.id,
+        tenantId: ExpectedData.tenantId,
```
zhang-shengping: the same as before

app/waf/src/controllers/declaration.controller.ts
```diff
@@ -333,4 +342,56 @@ export class DeclarationController extends BaseController {
...
+
+    let application = await this.applicationRepository.findById(applicationId);
+
+    if (!application.adcId) {
+      throw new HttpErrors.UnprocessableEntity(
+        'No target ADC to perform deploy action',
+      );
+    }
+
+    let adc = await this.adcRepository.findById(application.adcId);
```
zhang-shengping: do we need to select by tenantId here?

zongzw: BTY,  can we replace find with findById at #L367?

zhaoqin-github:
This change set is duplicated with PR #160. Let's commit all changes in #160 as a whole.

PR #151, opened 2019-05-24
add public flag for wafpolicy multi-tenant

zhang-shengping:


app/waf/src/controllers/wafpolicy.controller.ts
```diff
@@ -89,9 +89,12 @@ export class WafpolicyController extends BaseController {
   ): Promise<CollectionResponse> {
     return new CollectionResponse(
       Wafpolicy,
-      await this.wafpolicyRepository.find(filter, {
-        tenantId: await this.tenantId,
-      }),
+      await this.wafpolicyRepository.find(
+        {
```
zongzw: Add its test case?

zhang-shengping: we need multiple tenants in context and change them dynamically.

app/waf/src/controllers/wafpolicy.controller.ts
```diff
@@ -109,8 +112,13 @@ export class WafpolicyController extends BaseController {
   ): Promise<Response> {
     return new Response(
       Wafpolicy,
-      await this.wafpolicyRepository.findById(id, undefined, {
-        tenantId: await this.tenantId,
+      await this.wafpolicyRepository.findById(id, {
+        where: {
+          and: [
+            {or: [{tenantId: await this.tenantId}, {public: true}]},
```
zongzw: Just thinking that `id` can identify anything. So there may be no need to `and` such 2 conditions. What do you think?

app/waf/src/controllers/wafpolicy.controller.ts
zongzw: Why for that?

The test should be 
1. create the wafpolicy data whose property `public === true` and `tenantId === t1`.
2. access the api with `tenantId === t2` and check the return value is not empty and tenantId === t1. 

Is it?

app/waf/src/controllers/wafpolicy.controller.ts
zhang-shengping: I hope so, but it does not.

zongzw: why "does not" ?

zhang-shengping: it becomes somethig like "{{or: [{xxx: xxx}, {xxx:xxx}]}, {"and":[{xxx:xxx}]}}"

app/waf/test/acceptance/wafpolicy.controller.acceptance.ts
```diff
@@ -104,6 +104,49 @@ describe('WafpolicyController', () => {
     expect(toJSON(wafpolicy)).to.containDeep(response.body.wafpolicies[0]);
   });
 
+  it(
```
zongzw: Great.

app/waf/src/controllers/wafpolicy.controller.ts
zongzw: What I mean is if user passes an Id, we find the record with only 1 condition(id), no other condition.

zhang-shengping: 1. we want to get a public a wafpolicy whose tenantId is not the same as we used now.
2. if its without any other condition, we may get a wafpolicy which belongs to other tenant and is not public. or we can only get the wafpolicy which is in our own tenant (with the {tenantId: tenantId} option parameter).

zongzw: ... Got it. Thanks.

PR #153, opened 2019-05-27
Rename tdMgr to asgMgr

zhaoqin-github:


zhaoqin-github:
@zongzw I would post a similar comment to your last renaming pr.... This pr is to made the changes that you did not make in your last renaming change set. If your points is "we are necessary to have one asg service object, but we are not necessary to have one asg manager object", it will make others confused. Or reverting your last renaming pr is another option....

zongzw:
> @zongzw I would post a similar comment to your last renaming pr.... This pr is to made the changes that you did not make in your last renaming change set. If your points is "we are necessary to have one asg service object, but we are not necessary to have one asg manager object", it will make others confused. Or reverting your last renaming pr is another option....

Yeah, my mistake in last PR. My expect is to have on asg datasource, but different services&managers. 
I plan to revert last PR, and post a new one to change datasource name only. What do you think.

zhaoqin-github:
> > @zongzw I would post a similar comment to your last renaming pr.... This pr is to made the changes that you did not make in your last renaming change set. If your points is "we are necessary to have one asg service object, but we are not necessary to have one asg manager object", it will make others confused. Or reverting your last renaming pr is another option....
> 
> Yeah, my mistake in last PR. My expect is to have on asg datasource, but different services&managers.
> I plan to revert last PR, and post a new one to change datasource name only. What do you think.

So my mistake is merging your renaming pr too quickly....  Usually my personal thinking is to make code structure become simple, in order to let most persons understand it easier. If your new idea would be more complex than the current code structure, I might vote NO perhaps, but that will depend on your new code... Now, I would 60% select to merge this, 40% select revert your last renaming pr. Maybe you can take a little more time to think over your new idea, and post a new change set.

zongzw:
> 
> So my mistake is merging your renaming pr too quickly.... Usually my personal thinking is to make code structure become simple, in order to let most persons understand it easier. If your new idea would be more complex than the current code structure, I might vote NO perhaps, but that will depend on your new code... Now, I would 60% select to merge this, 40% select revert your last renaming pr. Maybe you can take a little more time to think over your new idea, and post a new change set.

The complex is relative, I think. 
I don't think my idea is new, also I don't think my idea is too complex to understand for you, for anyone. Only that you don't want to understand it. 
IF...ELSE is simple, but it doesn't solve our problem. As if you know design pattern, I think your view to it may be changed. 
To me now, it's clear, not because it's written by myself, but because it applies to the standard of designing. 

If you think my design is not a good solution, think of a better one to replace it.

zhaoqin-github:
Already merged by another commit 93affaeb3d350a059583d421b9ea67973862f524

PR #155, opened 2019-05-28
Save trusted device id

zhaoqin-github:


zhaoqin-github:
> In the future, How about organizing all these generated information(networks' portId, fixedIp, vmId, bigip's connection info) into a central place, such as, in adc.management.
> So that it can have no influence to user's initial inputs.

I am fine to save them in one object property. But we will need a way to operate them, eg. comparing management ip, in db layer. If filtering those properties is very easy, as regular db column, that will be great.

PR #156, opened 2019-05-29
add/del/update applications for one tenant

jikui:


zongzw:
@jikui @olvandeng The code is nearly 500 lines, could we have a meeting for reviewing it when it's ready? Maybe tomorrow afternoon...

app/waf/src/controllers/application.controller.ts
```diff
@@ -172,6 +177,160 @@ export class ApplicationController extends BaseController {
     });
   }
 
+  @post(prefix + '/applications/{applicationId}/append', {
```
zhaoqin-github: Why not reusing deploy/cleanup actions? I feel we may only provide two action APIs for end users, one is to deploy configuration, and the other is to clean the deployed configuration. End user may not need to know the difference between 'append' and 'replace'.

jikui:
> @jikui @olvandeng The code is nearly 500 lines, could we have a meeting for reviewing it when it's ready? Maybe tomorrow afternoon...

Not so much changes actually. The replace and append functions and their acceptance testing funcations are nearly the same.

app/waf/src/controllers/application.controller.ts
jikui: That's ok and will greatly simplify the codes...

jikui: I will make the changes accordingly.

jikui:
> Need a little more time to review. My 1st suggestion is to improve test coverage a little.

Added more test cases and the percentage raised +0.05% now.

app/waf/src/models/as3.model.ts
```diff
@@ -1,13 +1,69 @@
...
+    adc: Adc,
+    application: Application,
+    operation: patchOP,
+    declaration?: Declaration,
+  ) {
+    if (adc.management) {
+      this.targetHost = adc.management.ipAddress;
+      this.targetPort = adc.management.tcpPort;
+    }
+    //TODO: remove admin/pass after implement trusted connection
```
zhaoqin-github: We may use adc.management.username/password here. Temporarily hard code it here is also OK, since we will move to TrusedProxy soon.

app/waf/test/acceptance/application.controller.acceptance.ts
```diff
@@ -218,13 +218,68 @@ describe('ApplicationController', () => {
...
+  );
+
   it(
     'post ' + prefix + '/applications/{id}/deploy: deploy application',
     async () => {
       let adc = await givenAdcData(wafapp);
       let application = await givenApplicationData(wafapp, {
         adcId: adc.id,
       });
+      let applicationNew = await givenApplicationData(wafapp, {
```
zhaoqin-github: I am not able to understand the purpose of this test...

app/waf/test/acceptance/application.controller.acceptance.ts
```diff
@@ -218,13 +218,68 @@ describe('ApplicationController', () => {
       .expect(204);
   });
 
+  it(
+    'post ' +
+      prefix +
+      '/applications/{id}/deploy: deploy an unspported action.',
```
zhaoqin-github: What is "unspported action"? The AS3 action is made by our code. Why it will become "unspported?

jikui: correct the comments with "deploy an application without partition".

app/waf/test/acceptance/application.controller.acceptance.ts
jikui: The purpose is test the two types of exceptions, 1) the exception with error string "InvalidPatchOperationError: path does not existInvalidPatchOperationError: path does not exist", 2) the exception with error string that isn't "InvalidPatchOperationError: path does not exist".

PR #160, opened 2019-05-30
Add trustedproxy to deploy as3 objects.

zongzw:


app/waf/src/controllers/declaration.controller.ts
```diff
@@ -333,4 +342,60 @@ export class DeclarationController extends BaseController {
       .declarations(applicationId)
       .delete({and: [{id: declarationId}, {tenantId: await this.tenantId}]});
   }
+
+  // TODO: not sure how to name the API path. It's time for us to discuss all of them again.
+  @post(prefix + '/declarations/{declarationId}/deploy', {
```
zhaoqin-github: It conflicts with PR #149. And we need to vote resource uri first.

app/waf/src/models/as3.model.ts
```diff
@@ -41,4 +41,17 @@ export class AS3DeployRequest {
       });
     }
   }
+
+  static generateDeployBody(declaration: Declaration): object {
```
zhaoqin-github: Is it duplicated with the constructor?

app/waf/src/models/declaration.model.ts
```diff
@@ -29,3 +29,19 @@ export class Declaration extends CommonEntity {
     super(data);
   }
 }
+
+export class ASGDeployRequest extends Entity {
```
zhaoqin-github: Does it need to inherits from Entity?

app/waf/src/services/trustedproxy.service.ts
```diff
@@ -0,0 +1,55 @@
+import {getService} from '@loopback/service-proxy';
```
zhaoqin-github: Maybe we can merge it with asg.service.ts

app/waf/src/models/as3.model.ts
zongzw: If we don't use AS3, the AS3* staff will be obsolete. 
Also, I don't think generating a deployment body, `adc application` must be provided.
What's more, I just think that this part can be simplified to a function.

app/waf/src/models/declaration.model.ts
zongzw: It's used to generate the openapi schema, during which @property can decorate only Entity's properties. 

It will report error like this if no `extends Entity`: 
```src/controllers/declaration.controller.ts:359:9 - error TS2345: Argument of type 'typeof ASGDeployRequest' is not assignable to parameter of type 'typeof Entity'.
  Type 'typeof ASGDeployRequest' is missing the following properties from type 'typeof Entity': getIdOf, buildWhereForId, modelName, definition```

File: app/waf/src/services/trustedproxy.service.ts (hunk 4 above)
Comment (zongzw) [reply app/waf/src/services/trustedproxy.service.ts]: There are 2 reasons I didn't do that:
1. asg* staff is frequently coding changing. Coding it within that file will lead more merging effort. 
2. We can use asg as a datasource name, but the kinds of functionalities are provided by extensions. 
What functions asg provides depends on what extensions it has. I think separate the extensions into separate files can make us easier to decouple changes.

File: app/waf/src/models/as3.model.ts (hunk 2 above)
Comment (zhaoqin-github) [reply app/waf/src/models/as3.model.ts]: We still need the code to create AS3 request, but we just post the AS3 request via proxy.

File: app/waf/src/models/declaration.model.ts (hunk 3 above)
Comment (zhaoqin-github) [reply app/waf/src/models/declaration.model.ts]: Is it an action request body which includes ADC id to specify target device? If true, I got its meaning. Thanks!

File: app/waf/src/controllers/declaration.controller.ts (hunk 1 above)
Comment (zongzw) [reply app/waf/src/controllers/declaration.controller.ts]: Done, changed it.

File: app/waf/src/models/as3.model.ts (hunk 2 above)
Comment (zongzw) [reply app/waf/src/models/as3.model.ts]: Reused the AS3DeployRequest constructor.

File: app/waf/src/services/trustedproxy.service.ts (hunk 4 above)
Comment (zongzw) [reply app/waf/src/services/trustedproxy.service.ts]: Moved to ASGManager.

File: app/waf/src/controllers/declaration.controller.ts [hunk 5]
```diff
@@ -333,4 +342,69 @@ export class DeclarationController extends BaseController {
...
+
+        let mgmt = adc.management!;
+        let proxyMgr = await ASGManager.instanlize();
+
+        try {
+          let as3Request = new AS3DeployRequest(adc, application, declaration);
+          await proxyMgr.deploy(
+            mgmt.ipAddress,
+            mgmt.tcpPort,
+            as3Request.declaration,
```
Niklaus-xie: this is only passing as3Request.declaration. isn't the following also needed in the body?: e.g.   "class": "AS3",  "action": "deploy", etc.

zongzw: No need I think. So I'd prefer to remove all AS3 hints from the code. 

AS3DeployRequest may be useless, maybe...

PR #164, opened 2019-06-04
Remove the lastErr when next state comes.

zongzw:
(cherry picked from commit 6d85aa35acbed0ee26701f14534ba455832cdb2f)

### DIFF
--- a/app/waf/src/controllers/adc.controller.ts
+++ b/app/waf/src/controllers/adc.controller.ts
@@ -141,7 +141,7 @@ export class AdcController extends BaseController {
     };
 
     try {
-      await this.serialize(adc, {status: AdcState.TRUSTING});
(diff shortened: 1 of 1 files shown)


zongzw:
> Usually, the lastError is a record to tracking previous failure. It may not be remove, after state changes. What we need is to adding a timestamp for lastError.

It makes the response confused when the status and lastErr conflict, eg. ACTIVE, 'ONBOARDERROR: xxxxx'. 
For the "the lastError is a record to tracking previous failure", it should be in another place to record it. I think lastErr is the message telling the user what to do next.

Niklaus-xie:
LGTM  if the cherry-pick msg is removed. i don't see a cherry-pick coming.

zongzw:
Hold on this merging and guarantee the #170 to be merged.

PR #166, opened 2019-06-05
Remove PRODUCT_RELEASE, after that, all request need auth*.

zongzw:


app/waf/src/sequence.ts
```diff
@@ -83,10 +83,13 @@ export class MySequence implements SequenceHandler {
   }
 
   async authRequest(context: RequestContext) {
-    if (!process.env.PRODUCT_RELEASE) return;
-
     let {request} = context;
 
+    let ignoredAuthPaths = ['/openapi.json', '/explorer'];
+    for (let u of ignoredAuthPaths) {
+      if (request.path.startsWith(u)) return;
```
zhang-shengping: If client request this path: /openapi.json', '/explorer', client will get nothing?

zongzw: In that case, the client can get a response without auth.

zhaoqin-github: Should we allow 'ping' without auth token? Or, which api endpoint should we use for healthcheck?

zongzw: The `ping` calls as3 API. I think that we should not allow anyone to `ping` without auth*. 
The client who wants to ping should have the auth info which is used for later works.

zhaoqin-github: A healthcheck call may not need to include openstack user token. That is not the topic of this pr, let us do that later.

zongzw: yea, we can add it when necessary.

zongzw:
hold on, not to merge. let PR #170 be merged in first.

PR #169, opened 2019-06-05
Refresh and use user token's endpoints for openstack api call.

zongzw:
Why do we have this PR:

We are now using admin token to get service endpoints(compute, network, identity..)
The normal way should be using the user token. 2 benefits: 

1. The passed-in user token should not be used directly, since it may be expired in a very short time(eg. < 1s), but the operation takes a longer time. So we need to refresh a new user token and do the following works with it. 
2. We can get service endpoints directly from user token's catalogs, instead of assembling it from admin tokens.

### DIFF
--- a/app/waf/src/controllers/adc.controller.ts
+++ b/app/waf/src/controllers/adc.controller.ts
@@ -56,6 +56,7 @@ import {
   OnboardingManager,
   BigipBuiltInProperties,
   ASGServiceProvider,
+  AuthedToken,
 } from '../services';
 import {checkAndWait, merge} from '../utils';
 
@@ -417,7 +418,13 @@ export class AdcController extends BaseController {
     });
 
     let addonReq = {
-      userToken: await this.reqCxt.get(WafBindingKeys.Request.KeyUserToken),
+      userToken: await this.reqCxt
+        .get(WafBindingKeys.Request.KeyUserToken)
+        .then(token =>
+          this.wafapp
+            .get(WafBindingKeys.KeyAuthWithOSIdentity)
+            .then(authHelper => authHelper.solveUserToken(token)),
+        ),
       tenantId: await this.reqCxt.get(WafBindingKeys.Request.KeyTenantId),
     };
     this.adcStCtr = new AdcStateCtrlr(adc, addonReq);
@@ -525,7 +532,7 @@ export class AdcController extends BaseController {
         let userdata: string = await this.cUserdata(rootPass, adminPass);
 
         let serverParams: ServersParams = {
-          userTenantId: addon.tenantId,
+          userTenantId: addon.userToken!.tenantId,
           vmName: adc.id,
           imageRef: adc.compute.imageRef,
           flavorRef: adc.compute.flavorRef,
@@ -541,7 +548,7 @@ export class AdcController extends BaseController {
         };
 
         await computeHelper
-          .createServer(addon.userToken, serverParams)
+          .createServer(addon.userToken!, serverParams)
           .then(response => {
             adc.compute.vmId = response;
             adc.management = {
@@ -578,7 +585,7 @@ export class AdcController extends BaseController {
           if (net.fixedIp) portParams.fixedIp = net.fixedIp;
 
           await networkHelper
-            .createPort(addon.userToken, portParams)
+            .createPort(addon.userToken!, portParams)
             .then(async port => {
               net.fixedIp = port.fixedIp;
               net.macAddr = port.macAddr;
@@ -610,7 +617,10 @@ export class AdcController extends BaseController {
     let reclaimFuncs: {[key: string]: Function} = {
       license: async () => {
         let doMgr = await OnboardingManager.instanlize(this.wafapp);
-        let doBody = await doMgr.assembleDo(adc, {onboarding: false});
+        let doBody = await doMgr.assembleDo(
+          adc,
+          Object.assign(addon, {onboarding: false}),
+        );
         this.logger.debug(
           'Json used for revoke license: ' + JSON.stringify(doBody),
         );
@@ -642,7 +652,7 @@ export class AdcController extends BaseController {
 
           try {
             let portId = adc.networks[network].portId!;
-            await networkMgr.deletePort(addon.userToken, portId).then(() => {
+            await networkMgr.deletePort(addon.userToken!, portId).then(() => {
               this.logger.debug(`Deleted port ${portId}`);
               delete adc.networks[network].portId;
               delete adc.networks[network].fixedIp;
@@ -661,7 +671,11 @@ export class AdcController extends BaseController {
         );
         if (adc.compute.vmId) {
           await computeMgr
-            .deleteServer(addon.userToken, adc.compute.vmId!, addon.tenantId)
+            .deleteServer(
+              addon.userToken!,
+              adc.compute.vmId!,
+              addon.userToken!.tenantId,
+            )
             .then(() => {
               this.logger.debug(`Deleted the vm ${adc.compute.vmId!}`);
               delete adc.compute.vmId;
@@ -698,7 +712,10 @@ export class AdcController extends BaseController {
       await this.serialize(adc, {status: AdcState.ONBOARDING});
 
       let doMgr = await OnboardingManager.instanlize(this.wafapp);
-      let doBody = await doMgr.assembleDo(adc, {onboarding: true});
+      let doBody = await doMgr.assembleDo(
+        adc,
+        Object.assign(addon, {onboarding: true}),
+      );
       let doId = await doMgr.onboarding(doBody);
 
       await checkAndWait(() => doMgr.isDone(doId), 240)
@@ -716,8 +733,8 @@ export class AdcController extends BaseController {
 }
 
 export type AddonReqValues = {
-  userToken: string;
-  tenantId: string;
+  userToken?: AuthedToken;
+  onboarding?: boolean;
 };
 
 export class AdcStateCtrlr {
--- a/app/waf/src/datasources/openstack.datasource.json
+++ b/app/waf/src/datasources/openstack.datasource.json
@@ -41,6 +41,19 @@
       }
     },
     {
+      "description": "identity v2 refresh user's token.",
+      "template": {
+        "method": "POST",
+        "url": "{url}",
+        "responsePath": "$"
+      },
+      "functions": {
+        "v2RefreshToken": [
+          "url"
+        ]
+      }
+    },
+    {
       "description": "identity v3 get auth token with user/password",
       "template": {
         "method": "POST",
@@ -73,6 +86,22 @@
           "url",
           "adminToken",
           "userToken"
+        ]
+      }
+    },
+    {
+      "description": "identity v3 refresh user's token.",
+      "template": {
+        "method": "POST",
+        "url": "{url}",
+        "responsePath": "$",
+        "responseHeaders": true,
+        "body": "{body}"
+      },
+      "functions": {
+        "v3RefreshToken": [
+          "url",
+          "body"
         ]
       }
     },
--- a/app/waf/src/keys.ts
+++ b/app/waf/src/keys.ts
@@ -56,7 +56,7 @@ export namespace WafBindingKeys {
     export const KeyTenantId = BindingKey.create<string>(
       'context.request.tenantid',
     );
-    export const KeyUserToken = BindingKey.create<string>(
+    export const KeyUserToken = BindingKey.create<AuthedToken>(
       'context.request.usertoken',
     );
   }
--- a/app/waf/src/sequence.ts
+++ b/app/waf/src/sequence.ts
@@ -27,7 +27,7 @@ import {
   HttpErrors,
 } from '@loopback/rest';
 import {factory} from './log4ts';
-import {AuthWithOSIdentity, AuthedToken} from './services';
+import {AuthWithOSIdentity} from './services';
 import {CoreBindings} from '@loopback/core';
 import {WafApplication} from '.';
 import {WafBindingKeys} from './keys';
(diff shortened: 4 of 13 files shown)


zongzw:
> I do not know this part of code very well, but I have a general question: Is it appropriate for us to refresh user's keystone token? Will the previous token become invalid, after we refreshing it?

Refreshing a token means using the given token to retrieve a new token. The two tokens are different ones. And the later token is new with full expiration time.

Niklaus-xie:
the default time for a token is 1 hour? if i remember correctly. what's more, that is also configurable in keystone.  Have no idea why we need to refresh for a new token inside our code.  If a token provided is expired, shouldn't we ask for a valid token instead?

zhaoqin-github:
> > I do not know this part of code very well, but I have a general question: Is it appropriate for us to refresh user's keystone token? Will the previous token become invalid, after we refreshing it?
> 
> Refreshing a token means using the given token to retrieve a new token. The two tokens are different ones. And the later token is new with full expiration time.

Will the old token become invalid, when the new one is generated?

zongzw:
> the default time for a token is 1 hour? if i remember correctly. what's more, that is also configurable in keystone. Have no idea why we need to refresh for a new token inside our code. If a token provided is expired, shouldn't we ask for a valid token instead?

The expiration time is decided by keystone(default is 1 hour). 
The token user provides may expire in the middle of the operation(e.g. create port/virtual server..). We have no way to ask the user to input a new token, but fail and exit. 
Refreshing a token can guarantee the operation done, since we have accurate expire time thus(token is new).

zongzw:
> > > I do not know this part of code very well, but I have a general question: Is it appropriate for us to refresh user's keystone token? Will the previous token become invalid, after we refreshing it?
> > 
> > 
> > Refreshing a token means using the given token to retrieve a new token. The two tokens are different ones. And the later token is new with full expiration time.
> 
> Will the old token become invalid, when the new one is generated?

The old one is user's, it'll not become invalid, it is owned by user and has its own expire time.

Niklaus-xie:
if the token provided by the user expires in a short time period(e.g.  several seconds) , we should let it fail and reports an error   because refreshing for a new one  is out of our scope.  I don't think we need to refresh for a new one in order to ensure the success.

zongzw:
> if the token provided by the user expires in a short time period(e.g. several seconds) , we should let it fail and reports an error because refreshing for a new one is out of our scope. I don't think we need to refresh for a new one in order to ensure the success.

It sounds like more that it fails because our operation takes too long the time, not because user's token is nearly expired. 
User may say that "the token I give you is valid, why do you give me an error saying invalid token...". 

Refreshing the token, at the same time, is the way to validate the old token.

Niklaus-xie:
are you suggesting if any api of those project(e.g. nova, glance, heat etc) takes a relatively longer time to finish,  that api should include a refresh-token logic inside it because the token might expire ?       Anyway,  i think we should not bring any NOT-MUST-HAVE logic into our code and try to keep our code as simple as possible, so I still don't see the necessity for that.  Also, I doubt there is such kind of refresh-token logic in other openstack projects.  Go ahead and merge it if you still think you want this part.

zongzw:
> are you suggesting if any api of those project(e.g. nova, glance, heat etc) takes a relatively longer time to finish, that api should include a refresh-token logic inside it because the token might expire ? Anyway, i think we should not bring any NOT-MUST-HAVE logic into our code and try to keep our code as simple as possible, so I still don't see the necessity for that. Also, I doubt there is such kind of refresh-token logic in other openstack projects. Go ahead and merge it if you still think you want this part.

To me, I prefer to make the code as stable as possible when we notice the very issue, even only a possible issue.

zhaoqin-github:
> > > > I do not know this part of code very well, but I have a general question: Is it appropriate for us to refresh user's keystone token? Will the previous token become invalid, after we refreshing it?
> > > 
> > > 
> > > Refreshing a token means using the given token to retrieve a new token. The two tokens are different ones. And the later token is new with full expiration time.
> > 
> > 
> > Will the old token become invalid, when the new one is generated?
> 
> The old one is user's, it'll not become invalid, it is owned by user and has its own expire time.

I feel this idea is to generate a new token from a existing token provided by user, and the new token will be used by our app to manipulate openstack resources. If the existing token will not be impacted, that can be an idea. And we may also consider to revoke this new token appropriately, since it is a new resource created by us. Another thing which should be consider is the scope of the token. The new token should be a project scope token, because our app might not manipulate system scope resources. However, the token input by user may be a system scope token. For example, an admin who has a system scope token hope to modify an ADC in a user's project. We will need to generate a new token for this specific project scope from admin's token. I am not quite sure if keystone supports that.

zongzw:
> > > > > I do not know this part of code very well, but I have a general question: Is it appropriate for us to refresh user's keystone token? Will the previous token become invalid, after we refreshing it?
> > > > 
> > > > 
> > > > Refreshing a token means using the given token to retrieve a new token. The two tokens are different ones. And the later token is new with full expiration time.
> > > 
> > > 
> > > Will the old token become invalid, when the new one is generated?
> > 
> > 
> > The old one is user's, it'll not become invalid, it is owned by user and has its own expire time.
> 
> I feel this idea is to generate a new token from a existing token provided by user, and the new token will be used by our app to manipulate openstack resources. If the existing token will not be impacted, that can be an idea. And we may also consider to revoke this new token appropriately, since it is a new resource created by us. Another thing which should be consider is the scope of the token. The new token should be a project scope token, because our app might not manipulate system scope resources. However, the token input by user may be a system scope token. For example, an admin who has a system scope token hope to modify an ADC in a user's project. We will need to generate a new token for this specific project scope from admin's token. I am not quite sure if keystone supports that.

Currently, we only have project-scoped token. 
And currently, we have no scenario to need admin token to do something. 
Maybe in the future. What's good is: 
For admin token in our code, we already have a project scope admin token(refer to `resolvedAdminToken`). All of our API requests are in project scope.

zongzw:
> I see this code change will refresh user token for every api call. Is that right? If true, I think that will generate too much useless data in keystone. If the purpose of this change is to avoid token expiration problem during a time-cost operation. We may need to identify which operation does require to refresh token.

`generate too much useless data in keystone`
How much? Do you have a precise measurement on that? meaningless concern to this point, I think. 
Keystone would be broken down because of too many users login in public cloud?
Does WAFAAS will be trigger thousands of times in 1 sec? WAFAAS will crash earlier before keystone crash, is it right?

"We may need to identify which operation does require to refresh token." how to identify? 
Do you know how much complexity does it add to our code?

zhaoqin-github:
> Why do we have this PR:
> 
> We are now using admin token to get service endpoints(compute, network, identity..)
> The normal way should be using the user token. 2 benefits:
> 
> 1. The passed-in user token should not be used directly, since it may be expired in a very short time(eg. < 1s), but the operation takes a longer time. So we need to refresh a new user token and do the following works with it.
> 2. We can get service endpoints directly from user token's catalogs, instead of assembling it from admin tokens.

To my understanding this change set hope to address two different topics:

1. Using user token instead of admin token to interact with OpenStack.

2. Refresh user token, if we really need to refresh it.

For topic 1, I will vote 'yes'. That seems to be a bug.

For topic 2, it seems to be an nice-to-have enhancement. And its code change might be a little more complex the the code change for topic 2.

If you can propose two separated change set at the beginning, that would be a little better. We may review and approve the code change for topic 1 a little easier.

And my questions for topic 2 is which API will require us to refresh keystone token. Maybe we will have 3 cases:

1. If an API handler does not interact with OpenStack, it only need to validate the tenant information of existing user token. That means it never need to refresh the token. 

2. If an API handler only need to call OpenStack API once, and will do that immediately, it needn't to refresh the token.

3. If an API handler need to call OpenStack API several times, and there will be big intervals between those calls, it may need to refresh the token, if the token is very close to be expired. We know such kind of problems a lot in ancient years. Since nova or neutron did not improve it, we would see some "token expired" problem in stress testing. I am glad to see someone would improve it, but we may need to scope this improved behavior under specific circumstances.

zongzw:
> > Why do we have this PR:
> > We are now using admin token to get service endpoints(compute, network, identity..)
> > The normal way should be using the user token. 2 benefits:
> > 
> > 1. The passed-in user token should not be used directly, since it may be expired in a very short time(eg. < 1s), but the operation takes a longer time. So we need to refresh a new user token and do the following works with it.
> > 2. We can get service endpoints directly from user token's catalogs, instead of assembling it from admin tokens.
> 
> To my understanding this change set hope to address two different topics:
> 
>     1. Using user token instead of admin token to interact with OpenStack.
> 
>     2. Refresh user token, if we really need to refresh it.
> 
> 
> For topic 1, I will vote 'yes'. That seems to be a bug.
> 
> For topic 2, it seems to be an nice-to-have enhancement. And its code change might be a little more complex the the code change for topic 2.
> 
> If you can propose two separated change set at the beginning, that would be a little better. We may review and approve the code change for topic 1 a little easier.
> 
> And my questions for topic 2 is which API will require us to refresh keystone token. Maybe we will have 3 cases:
> 
>     1. If an API handler does not interact with OpenStack, it only need to validate the tenant information of existing user token. That means it never need to refresh the token.
> 
>     2. If an API handler only need to call OpenStack API once, and will do that immediately, it needn't to refresh the token.
> 
>     3. If an API handler need to call OpenStack API several times, and there will be big intervals between those calls, it may need to refresh the token, if the token is very close to be expired. We know such kind of problems a lot in ancient years. Since nova or neutron did not improve it, we would see some "token expired" problem in stress testing. I am glad to see someone would improve it, but we may need to scope this improved behavior under specific circumstances.

You even have the wrong understanding to the original code and this change. How can you review the code? 
We never use admin token to operate on OS, can you please check from the code again. 
How can we use admin token to operate on OS and how can you endure that if we do it?

To the issue this PR locates, I consider it as a bug and you think that's an enhancement. 
I don't think it's an enhancement. It's an indeed existing possible problem because we have several sub operations in our API call. If the token expires in the middle of operations, it will make chaos that we can not easy to handle. 

Neutron nova do solve it but we do, I don't understand where's the problem? 
I also don't understand why do we need to compare their shortages and lean to be consistance with that?

At last, could you please check, check and make the code understood before you start the review?
How can we dare to use admin token to do operations? 
All the functions parameters are using 'userToken', haven't you seen it?

zhaoqin-github:
> > > Why do we have this PR:
> > > We are now using admin token to get service endpoints(compute, network, identity..)
> > > The normal way should be using the user token. 2 benefits:
> > > 
> > > 1. The passed-in user token should not be used directly, since it may be expired in a very short time(eg. < 1s), but the operation takes a longer time. So we need to refresh a new user token and do the following works with it.
> > > 2. We can get service endpoints directly from user token's catalogs, instead of assembling it from admin tokens.
> > 
> > 
> > To my understanding this change set hope to address two different topics:
> > ```
> > 1. Using user token instead of admin token to interact with OpenStack.
> > 
> > 2. Refresh user token, if we really need to refresh it.
> > ```
> > 
> > 
> > For topic 1, I will vote 'yes'. That seems to be a bug.
> > For topic 2, it seems to be an nice-to-have enhancement. And its code change might be a little more complex the the code change for topic 2.
> > If you can propose two separated change set at the beginning, that would be a little better. We may review and approve the code change for topic 1 a little easier.
> > And my questions for topic 2 is which API will require us to refresh keystone token. Maybe we will have 3 cases:
> > ```
> > 1. If an API handler does not interact with OpenStack, it only need to validate the tenant information of existing user token. That means it never need to refresh the token.
> > 
> > 2. If an API handler only need to call OpenStack API once, and will do that immediately, it needn't to refresh the token.
> > 
> > 3. If an API handler need to call OpenStack API several times, and there will be big intervals between those calls, it may need to refresh the token, if the token is very close to be expired. We know such kind of problems a lot in ancient years. Since nova or neutron did not improve it, we would see some "token expired" problem in stress testing. I am glad to see someone would improve it, but we may need to scope this improved behavior under specific circumstances.
> > ```
> 
> You even have the wrong understanding to the original code and this change. How can you review the code?
> We never use admin token to operate on OS, can you please check from the code again.
> How can we use admin token to operate on OS and how can you endure that if we do it?
> 
> To the issue this PR locates, I consider it as a bug and you think that's an enhancement.
> I don't think it's an enhancement. It's an indeed existing possible problem because we have several sub operations in our API call. If the token expires in the middle of operations, it will make chaos that we can not easy to handle.
> 
> Neutron nova do solve it but we do, I don't understand where's the problem?
> I also don't understand why do we need to compare their shortages and lean to be consistance with that?
> 
> At last, could you please check, check and make the code understood before you start the review?
> How can we dare to use admin token to do operations?
> All the functions parameters are using 'userToken', haven't you seen it?

Hi, you pr description said the code uses admin token...

zongzw:
> > > > Why do we have this PR:
> > > > We are now using admin token to get service endpoints(compute, network, identity..)
> > > > The normal way should be using the user token. 2 benefits:
> > > > 
> > > > 1. The passed-in user token should not be used directly, since it may be expired in a very short time(eg. < 1s), but the operation takes a longer time. So we need to refresh a new user token and do the following works with it.
> > > > 2. We can get service endpoints directly from user token's catalogs, instead of assembling it from admin tokens.
> > > 
> > > 
> > > To my understanding this change set hope to address two different topics:
> > > ```
> > > 1. Using user token instead of admin token to interact with OpenStack.
> > > 
> > > 2. Refresh user token, if we really need to refresh it.
> > > ```
> > > 
> > > 
> > > For topic 1, I will vote 'yes'. That seems to be a bug.
> > > For topic 2, it seems to be an nice-to-have enhancement. And its code change might be a little more complex the the code change for topic 2.
> > > If you can propose two separated change set at the beginning, that would be a little better. We may review and approve the code change for topic 1 a little easier.
> > > And my questions for topic 2 is which API will require us to refresh keystone token. Maybe we will have 3 cases:
> > > ```
> > > 1. If an API handler does not interact with OpenStack, it only need to validate the tenant information of existing user token. That means it never need to refresh the token.
> > > 
> > > 2. If an API handler only need to call OpenStack API once, and will do that immediately, it needn't to refresh the token.
> > > 
> > > 3. If an API handler need to call OpenStack API several times, and there will be big intervals between those calls, it may need to refresh the token, if the token is very close to be expired. We know such kind of problems a lot in ancient years. Since nova or neutron did not improve it, we would see some "token expired" problem in stress testing. I am glad to see someone would improve it, but we may need to scope this improved behavior under specific circumstances.
> > > ```
> > 
> > 
> > You even have the wrong understanding to the original code and this change. How can you review the code?
> > We never use admin token to operate on OS, can you please check from the code again.
> > How can we use admin token to operate on OS and how can you endure that if we do it?
> > To the issue this PR locates, I consider it as a bug and you think that's an enhancement.
> > I don't think it's an enhancement. It's an indeed existing possible problem because we have several sub operations in our API call. If the token expires in the middle of operations, it will make chaos that we can not easy to handle.
> > Neutron nova do solve it but we do, I don't understand where's the problem?
> > I also don't understand why do we need to compare their shortages and lean to be consistance with that?
> > At last, could you please check, check and make the code understood before you start the review?
> > How can we dare to use admin token to do operations?
> > All the functions parameters are using 'userToken', haven't you seen it?
> 
> Hi, you pr description said the code uses admin token...

```admin token to get service endpoints(compute, network, identity..)```, use admin token for what? have you read it carefully?

PR #170, opened 2019-06-07
Fix issue create/setup process with using AdcStateCtrler

zongzw:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -98,9 +104,9 @@ export class AdcController extends BaseController {
         );
       }
 
-      this.trustAdc(adc).then(() => {
+      await this.trustAdc(adc).then(async () => {
```
zhaoqin-github: May not need to wait here. POST request need to return, otherwise client may timeout.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -307,6 +304,7 @@ export class AdcController extends BaseController {
     @param(Schema.pathParameter('adcId', 'ADC resource ID')) id: string,
   ): Promise<void> {
     let adc = await this.adcRepository.findById(id, undefined, {
+      //await this.adcRepository.findById(id, undefined, {
```
zhaoqin-github: Remove this?

app/waf/src/controllers/adc.controller.ts
zongzw: This session of code(HW) is just used to debug code, will eventually be removed. 

With await, it's easy to debug forwards. It's OK to remove it, no impact to me.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -720,74 +670,154 @@ export class AdcController extends BaseController {
...
+      check: (ctrl: AdcStateCtrlr) => {
+        return Promise.resolve(true);
+      },
       next: [AdcState.POWERON],
     },
     {
+      failure: AdcState.POWERERR,
       state: AdcState.POWERON,
-      check: this.accessible(),
+      check: this.accessible,
```
zhaoqin-github: I hope to know how can we implement error exit path in this structure. For instance, after creating a VE instance in OpenStack, if the instance is ACTIVE and not pingable, we need to wait and check again; if the instance become ERROR, we need to stop waiting and move ADC state to POWERERROR. How does the code do this?

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: Whatever removing or not, that is wrong implementation. We may not return response after install AS3 rpm. That takes about 30 seconds.

zongzw: When removing `await`, there's something wrong to acceptance test. 

```
    1) post /adcaas/v1/adcs: create ADC HW with trust timeout
(node:94655) UnhandledPromiseRejectionWarning: Error: Entity not found: Adc with id "76e59636-845e-4d12-b163-77e03af3bb13"
    at AdcRepository.findById (/Users/zong/GitRepos/zongzw/f5-adcaas-openstack/app/waf/src/repositories/common.ts:52:13)
    at processTicksAndRejections (internal/process/next_tick.js:81:5)
(node:94655) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 14)
(node:94655) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

```

I give up on changing it back to the intermediate state.

zhaoqin-github: The current testing code in master brach supports this HW path. If your new change breaks the test, it will be your responsibility to to fix it. Or, if there is really no person need HW path to debug, we may delete this path, and re-enable HW path in the future.

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: I am just curious if you already have an idea to support error exiting or not...

app/waf/src/controllers/adc.controller.ts
zongzw: What does it mean "Whatever removing or not, that is wrong implementation. We may not return response after install AS3 rpm. ". Do you mean we should not await?

zongzw: > The current testing code in master brach supports this HW path. If your new change breaks the test, it will be your responsibility to to fix it. Or, if there is really no person need HW path to debug, we may delete this path, and re-enable HW path in the future.

I prefer to remove it, however, I'm not sure you or others still needs it. 
The related code is big(because it includes many test code). You can post a PR to remove it if necessary. 

The VE onboarding path's time cost is ~3 min. For the second/third/.. time is <1min. Should be convenient for testing already.

app/waf/src/controllers/adc.controller.ts
zongzw: We've got an initial idea on this, and @haoshen61  is working on this.

zhaoqin-github: If current code structure can not support it, let us do it in next pr.

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: I am fine to remove it. You may let others know that. Maybe Jikui or someone still use it.

app/waf/src/controllers/adc.controller.ts
zongzw: OK.

app/waf/src/controllers/adc.controller.ts
zongzw: @jikui Maybe creating a standalone API for the developing is more agile than placing them in create application of 'HW'.

jikui: > agile

Let's discuss this later when necessary.

zhaoqin-github: Let's make this HW path unchanged.

zongzw: Do you persist on this? 
I think we'll soon remove this `if`. -- HW. You also agree on this point. 

I don't think doing an investigation on it is valuable.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -732,74 +685,154 @@ export class AdcController extends BaseController {
...
+    try {
+      this.logger.debug('start to do onbarding');
+      await this.serialize(adc, {status: AdcState.ONBOARDING});
+
+      let doMgr = await OnboardingManager.instanlize(this.wafapp);
+      let doBody = await doMgr.assembleDo(adc, {onboarding: true});
+      let doId = await doMgr.onboarding(doBody);
+
+      await checkAndWait(() => doMgr.isDone(doId), 240)
+        .then(() =>
```
Niklaus-xie: seems to check whether DO is DONE has 2 steps ,  IMHO, it's better if we could do this consistently for other parts, meaning  all checks for DO also go in to the AdcStateEntry.check() method, which i think it's the intended design in that AdcStateCtrlr class.  But I don't mind if we leave it until next time.

zongzw: You are right. It's OK to putting them into state controller. 

For this case, `isDone` is to check the API execution is over. 
If it returns non-200, check will be terminated immediately. left it here may be earlier to discover error when happens.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -732,74 +685,154 @@ export class AdcController extends BaseController {
...
+      next: [AdcState.RECLAIMED],
+    },
   ];
 
   constructor(private adc: Adc, private addon: AddonReqValues) {}
 
   async readyTo(state: string): Promise<boolean> {
     let stateEntry = this.getStateEntry(this.adc.status);
-    return (await stateEntry.check) && stateEntry.next.includes(state);
+
```
Niklaus-xie: pls re-check if   readyTo() logic is correct.  think about the following scenario: e.g. call the {"create": null} or other apis twice?  it seems to me this readyTo() would return true in line 770, which does not seem to be correct.   pls check if running those for the 2nd time does not break anything.

zongzw: That's case which has been considered in adc.controller.ts:

```
420     if (adc.status.endsWith('ING'))
421       throw new HttpErrors.UnprocessableEntity(
422         `Adc status is ' ${adc.status}. Cannot be operated on, please wait for its finish.`,
423       );
```

In the future, this check should also be moved into state controller.

Niklaus-xie: It's not about the ING.   the scenario  i meant was:  when the 1st api call  already finished successfully (which means it is not anything ending with ING),   then  the  same api is called for the 2nd time, we  probably want to make sure calling for the 2nd time does not bring anything unexpected.

zongzw: Oops, it's indeed an issue. Done. please review.

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: We already talk that in the meeting... If you can make CI pass without modify this part of code, that should be better, because it will not impact others. I used to believe you can not make CI pass with your new refactoring change set, but you said you can....

zongzw: How about removing the HW path directly. I think no one is using it for now.

zongzw: It's no value to fix the code which is going away.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -732,74 +689,155 @@ export class AdcController extends BaseController {
...
+        }
+        return true;
+      })(),
+    );
+  }
+
+  private async trusted(ctrl: AdcStateCtrlr): Promise<boolean> {
+    if (!ctrl.adc.trustedDeviceId!) return false;
+
+    let asgMgr = await ctrl.getAsgMgr();
```
zhaoqin-github: let asgMgr = ctrl.asgMgr;

app/waf/src/controllers/adc.controller.ts
```diff
@@ -817,34 +855,28 @@ class AdcStateCtrlr {
...
       .catch(() => Promise.reject(false));
   }
 
-  private async getBigipMgr(): Promise<BigIpManager> {
-    if (!this.adc.management)
-      throw new Error(
-        `The management session of ADC is empty, cannot initialize bigip manager.`,
-      );
+  private async installed(ctrl: AdcStateCtrlr): Promise<boolean> {
+    let asgMgr = await ctrl.getAsgMgr();
```
zhaoqin-github: let asgMgr = ctrl.asgMgr;

app/waf/src/controllers/adc.controller.ts
```diff
@@ -732,74 +689,155 @@ export class AdcController extends BaseController {
...
+      username: mgmt.username,
+      password: mgmt.password,
+      ipAddr: mgmt.ipAddress,
+      port: mgmt.tcpPort,
+    });
   }
 
-  private async trusted(): Promise<boolean> {
-    return Promise.resolve(true);
+  private async getAsgMgr(): Promise<ASGManager> {
```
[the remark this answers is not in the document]
zongzw: What does it mean `asgMgr is a member of this class`? asgMgr is just a local variable, not a member of AdcStateCtrlr.

app/waf/src/controllers/adc.controller.ts
zongzw: It's not a member of ctrl.

app/waf/src/controllers/adc.controller.ts
zongzw: ditto.

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: Sorry, I used to assume the ctrl is adc controller.

PR #173, opened 2019-06-10
Create newman test

zhaoqin-github:


test/.gitignore
```diff
@@ -0,0 +1 @@
+newman
```
zongzw: I suggest to organize the e2e test into "e2e" folder, so that `acceptance` / `unit` / `e2e` test can be well managed under "test", also in future we may have multiple e2e test process. Place all of them under "test/" may get mixed-up.

test/onboard-collection.json
```diff
@@ -0,0 +1,1495 @@
+{
```
zongzw: 1. get-keystone-token: need to extract username password tenent id from body to environments.
    2. All the api request: add tenant-id in header.
    3. In the future, create network and related resources from OS and delete them after test done.
    4. For reusing pre-request script or test script, we can refer to here in future: https://stackoverflow.com/questions/41609009/is-it-possible-to-reuse-postman-scripts-among-different-apis-test-cases
    5. create-member: the MEMBER_IP should not be from environment but generated by pre-request itself or request from neutron. So do other IPs.
    6. create-wafpolicy: "url": "https://gitlab.com/zhaoqin/misc/raw/master/waf_demo.xml" can be extracted as an environment.

test/os-queens-environment.json
```diff
@@ -0,0 +1,169 @@
+{
+	"id": "8e0a9736-c3eb-452e-ae9d-64fd8e94cb0c",
```
zongzw: 1. Maybe we need to remove the generated environments: RULE_ID SERVICE_ID ….
2. Suggest that those generated environments be named with lower-case in collection.json so that we can distinguish them with  those given outside.

zhaoqin-github: When we run this test in postman, those var will be add to environment json again, and will be exported. So that everyone need to remove them before check-in file. I think we may not let other to take time to do that.

If all vars are in this json, the ones who do have intial value always require a initial value.

test/.gitignore
zhaoqin-github: It is under test folder of top directory, so it is not mixed with the unit test of each app.

zongzw: ok

test/onboard-collection.json
zhaoqin-github: 1: have been done.
2: not sure what does it mean.
3: Maybe or maybe not. For example, I am not sure if we will be able to create workable external network every time when we execute test script. Need to evaluate its doability.
4: We can evaluate it in the future.
5: It can be a todo.
6: We can change this right now.

zongzw: 2. 
With keystone v2.0, it requires to api caller to provide project info to authorize a token. While in v3 it doesn't. You are using v3, so there's no problem. But if you are using v2.0. Unauthorized will be reported from our code..

```
  let userToken: AuthedToken;
    try {
      userToken = await this.authWithOSIdentity
        .validateUserToken(<string>hdrToken)
        .then(authedObj => {
          this.logger.debug(`Authenticated OK: Request ID(${context.name})`);
          //this.logger.debug(JSON.stringify(authedObj));
          return authedObj;
        });
    } catch (error) {
      throw new HttpErrors.Unauthorized(error.message);
    }

    if (request.headers['tenant-id']) {        <= must be provided when keystone v2.0
      userToken.tenantId = <string>request.headers['tenant-id'];
    }
    if (!userToken.tenantId || userToken.tenantId === '') {
      throw new HttpErrors.BadRequest('BadRequest: tenant id is not provided');
    }

    context.bind(WafBindingKeys.Request.KeyTenantId).to(userToken.tenantId);
    context.bind(WafBindingKeys.Request.KeyUserToken).to(userToken.token);
  }
```

zhaoqin-github: Ok, I see. That is a way to become compatible with keystone v2 and v3 both, right? If true, let's change it right now.

zongzw: 3. 
In fact, external network is no different with others. It just refers to a normal network which can be created via neutron api. 
Later, we still need to create the client/server connection test sandbox in external/internal network. 
Currently to me, all of this process can be done with script/api call. That can make less our dependency to ourside OS.(To run e2e, we needn't to provide C/S host, networks separately.)

zongzw: > Ok, I see. That is a way to become compatible with keystone v2 and v3 both, right? If true, let's change it right now.

Yes, thus, if we let api calls are in unified form, our code can be simplified.

PR #174, opened 2019-06-10
use asg instead of as3

Niklaus-xie:


app/waf/src/controllers/application.controller.ts
```diff
@@ -62,7 +59,7 @@ export class ApplicationController extends BaseController {
     public declarationRepository: DeclarationRepository,
     @repository(AdcRepository)
     public adcRepository: AdcRepository,
-    @inject('services.AS3Service') public as3Service: AS3Service,
+    @inject('services.ASGService') public asgService: ASGService,
```
zongzw: The `asgService` in this class is not used, can be removed.

app/waf/src/controllers/application.controller.ts
```diff
@@ -278,8 +278,11 @@ export class ApplicationController extends BaseController {
       tenantId: tenantId,
     });
 
+    let mgmt = adc.management!;
+    let asgManager = await ASGManager.instanlize();
+
     let operation = patchOP.Remove;
     let req = new AS3PatchReqeust(adc, application, operation);
-    return await this.as3Service.deploy(AS3_HOST, AS3_PORT, req);
+    await asgManager.deploy(mgmt.ipAddress, mgmt.tcpPort, req);
```
zongzw: This call to deploy may need to be surrounded with try..catch..422 or 4xx, or it may response 500 error. Same case for `deployById` maybe.

app/waf/src/controllers/application.controller.ts
Niklaus-xie: it's actually used in   tests . so i modified tests code to get rid of it.

app/waf/src/controllers/application.controller.ts
Niklaus-xie: for deployById, it seems that's because it needs to consider when the partition is not there at all, and will try to use a full-deploy instead to try again. does not seem to need that part here.

app/waf/src/controllers/application.controller.ts
zongzw: good.

app/waf/src/controllers/application.controller.ts
zongzw: Do you mean this part of code will be replaced with some else? So no need to consider it for now?

Niklaus-xie: what i meant is that in the deployById(), they added the try/catch because they need to try "deploy-method" when "patch-method" fails(in case the partition is not found on bigip), we don't seem to need this here.

Niklaus-xie:
Did a rebase on master branch.

app/waf/src/controllers/application.controller.ts
zongzw: Just f2f talk, and got the core, in the future, we may need to refine the exception catching in several places of the code.

app/waf/src/controllers/application.controller.ts
```diff
@@ -220,6 +216,9 @@ export class ApplicationController extends BaseController {
       tenantId: tenantId,
     });
 
+    let mgmt = adc.management!;
```
zhaoqin-github: If you look at ADC controller, you will see that we also has a way to implicitly init ASG manager during controller object constructing. So that we will not have to explicitly create ASG manager object in every handler function.

Niklaus-xie: I actually intended to do that earlier, but it seems it didn't worked. Anyway, that's a different way to implement, I can try again later and will refactor that in another patch.

Niklaus-xie:
Removing some unnecessary AS3 logic also seems fine. I can do that in another PR and leave this one only for implementation because seems this one will have conflicts/interactions with jikui's user story. I would like to avoid some of those conflicts by merging this as early as possible and thus avoid some trouble of dependency issues.

app/waf/src/controllers/application.controller.ts
Niklaus-xie: now i remember, because asgService is not needed here and thus has already been removed from the constructor, so I now can not use the way which is used in AdcController now. So i will keep using this way.

zhaoqin-github: Ok. It is up to you

PR #178, opened 2019-06-14
Assert service virtualAddress availability and then bind to adc ext port. 

zongzw:


zongzw:
> I am not sure why this branch includes 3 commits, and the 1st and 2nd commits are duplicated and not related to this topic... Can you please fix your branch first?

It bases on PR #169  which should be reviewed first.

zhaoqin-github:
> > I am not sure why this branch includes 3 commits, and the 1st and 2nd commits are duplicated and not related to this topic... Can you please fix your branch first?
> 
> It bases on PR #169 which should be reviewed first.

Hi, that is a wrong way to propose code changes. #169 is another topic which is not a dependency of VIP setting topic. You should make two  separately change set for them, in order to have them be reviewed individually.

app/waf/src/controllers/declaration.controller.ts
```diff
@@ -423,4 +432,85 @@ export class DeclarationController extends BaseController {
...
+          `not found the external network from adc: ${adc.networks}`,
+        );
+      })();
+
+      let netHelper = await this.wafapp.get(WafBindingKeys.KeyNetworkDriver);
+
+      let userToken = await this.reqCxt.get(
+        WafBindingKeys.Request.KeyUserToken,
+      );
+      let ports = await netHelper.getPortsInfo(userToken, netId);
```
Niklaus-xie: getting all ports of this network then checking each one is too heavy. we need to find a better way.

app/waf/src/controllers/declaration.controller.ts
```diff
@@ -408,11 +415,13 @@ export class DeclarationController extends BaseController {
 
         try {
           let as3Request = new AS3DeployRequest(adc, application, declaration);
-          await proxyMgr.deploy(
-            mgmt.ipAddress,
-            mgmt.tcpPort,
-            as3Request.declaration,
+          let declBody = as3Request.declaration;
+
+          await this.assertVirtualAddressAvailability(adc, declBody).then(() =>
```
Niklaus-xie: one possible proposal: how about bindVirtualAddressToExt() directly, and catch the possible exception from neutron, if the addr is indeed occupied already. (instead of using the heavy assertVirtualAddressAvailability).

app/waf/src/controllers/declaration.controller.ts
zongzw: Or do you know how to query a port with query parameter of fixed_ips array?

app/waf/src/controllers/declaration.controller.ts
```diff
@@ -423,4 +432,85 @@ export class DeclarationController extends BaseController {
...
+      let ports = await netHelper.getPortsInfo(userToken, netId);
+
+      let addresses = findByKey(declaration, 'virtualAddresses');
+      if (addresses.length === 0)
+        throw new Error('not found virtual servers, absurd declaration.');
+      for (let addrset of addresses) {
+        for (let addr of <string[]>addrset) {
+          // check addr on netId is available.
+
+          for (let port of ports) {
```
Niklaus-xie: also seems to me there is a loophole here? is it possible that the address is already occupied by another tenant because this is only checking against this tenant's ports? (scenario:  network shared by many tenants)

zongzw: good point.

app/waf/src/controllers/declaration.controller.ts
zongzw: yeah, binding directly is better.

app/waf/src/controllers/declaration.controller.ts
Niklaus-xie: if my proposal is to be adopted, we don't need to "query a port with parameter of fixed_ips array" anymore. Also, we might be able to delete many codes, e.g. getPorts info etc.

PR #179, opened 2019-06-14
create the partition soon after installing the as3

jikui:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -855,6 +886,16 @@ export class AdcStateCtrlr {
     });
   }
 
+ private async partitioned(ctrl: AdcStateCtrlr): Promise<boolean> {
+    let bigipMgr = await ctrl.getBigipMgr();
+    try {
+      await bigipMgr.checkPartition(as3Name(ctrl.adc.tenantId));
+    } catch (error) {
+      return false;
```
zongzw: Maybe this.logger.debug to log the error is useful.

app/waf/src/models/as3.model.ts
```diff
@@ -114,3 +114,43 @@ export class AS3DeployRequest {
     }
   }
 }
+
+export class AS3PartitionRequest {
```
zongzw: Can it be merged with AS3DeployRequest?

app/waf/src/controllers/adc.controller.ts
```diff
@@ -855,6 +886,16 @@ export class AdcStateCtrlr {
     });
   }
 
+ private async partitioned(ctrl: AdcStateCtrlr): Promise<boolean> {
+    let bigipMgr = await ctrl.getBigipMgr();
+    try {
+      await bigipMgr.checkPartition(as3Name(ctrl.adc.tenantId));
```
zongzw: if `checkPartition` will return false?

PR #181, opened 2019-06-14
create the partition soon after installing the as3

jikui:


app/waf/src/services/bigip.service.ts
```diff
@@ -196,6 +196,18 @@ export class BigIpManager {
     );
   }
 
+  async checkPartition(partition: string): Promise<boolean> {
```
zongzw: I think the codes in bigip.service.ts is only used for retrieving information from bigip, has no responsibility of check...
So returning the gotten information back and left the upper caller to do the check is reasonable.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -424,7 +445,9 @@ export class AdcController extends BaseController {
 
     if (adc.status.endsWith('ING'))
       throw new HttpErrors.UnprocessableEntity(
-        `Adc status is ' ${adc.status}. Cannot be operated on, please wait for its finish.`,
+        `Adc status is ' ${
```
zongzw: You may need to delete `node_modules` folder and `npm install` again. tslint is updated, this line should not have code change.

app/waf/src/services/bigip.service.ts
jikui: Let us discuss this further. In my opinion, checkPartition is the function used to fit in your design.

Thanks.

app/waf/src/controllers/adc.controller.ts
jikui: Get it and will following this suggestion accordingly.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -243,6 +245,25 @@ export class AdcController extends BaseController {
     }
   }
 
+  async installPartition(adc: Adc): Promise<void> {
+    // Install partition after installing the AS3 agent.
+    let tenantName = adc.getAS3Name();
+    let mgmt = adc.management!;
+    let paritionObj = new AS3PartitionRequest(adc);
+    try {
+      await this.serialize(adc, {status: AdcState.PARTITIONING});
```
Niklaus-xie: also add lastErr to keep consistency?

app/waf/src/controllers/adc.controller.ts
```diff
@@ -243,6 +245,25 @@ export class AdcController extends BaseController {
...
+      await this.asgMgr.deploy(mgmt.ipAddress, mgmt.tcpPort, paritionObj);
+      await this.serialize(adc, {status: AdcState.ACTIVE});
+    } catch (err) {
+      this.logger.debug(`Creating partition ${tenantName} Error.`);
```
Niklaus-xie: you might want 2 use logger.error instead.

app/waf/test/acceptance/adc.controller.acceptance.ts
```diff
@@ -1104,7 +1104,7 @@ describe('AdcController test', () => {
             .set('tenant-id', ExpectedData.tenantId)
             .expect(200);
 
-          return resp.body.adc.status === 'ACTIVE';
+          return resp.body.adc.status === 'INSTALLED';
```
Niklaus-xie: after adding this new PARTITIONED state, this case now seems strange because we didn't add the PARTITION step for HW.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -479,8 +501,9 @@ export class AdcController extends BaseController {
     // install as3
     if (await this.adcStCtr.readyTo(AdcState.INSTALLED))
       await this.installAS3(adc);
-
     // create tenant
+    if (await this.adcStCtr.readyTo(AdcState.PARTITIONED))
+      await this.installPartition(adc);
```
Niklaus-xie: might want to consider the HW path around line#122.  we might want to make a decision about how to deal with HW now because it does not seem to install the partition for HW. TO me, HW path can be deleted.

app/waf/test/acceptance/adc.controller.acceptance.ts
jikui: outdated. The previous change just aims to please the UT.

app/waf/src/controllers/adc.controller.ts
jikui: done

app/waf/test/acceptance/adc.controller.acceptance.ts
jikui: changed it back.

app/waf/src/services/bigip.service.ts
jikui: resolved and moved the check logic outside.

app/waf/src/controllers/adc.controller.ts
jikui: corrected.

app/waf/src/controllers/adc.controller.ts
jikui: This is a legacy issue related to the HW design. Let's try to change this after the final decision is  done.

Niklaus-xie: ok. Another discussion and a future PR seems fine to me .

Niklaus-xie:
my major concerns are addressed. One remaining concern for us: HW path.

PR #186, opened 2019-06-17
Find ep from usertoken

zongzw:


zongzw:
> During functional test, I notice in our Queens env, a tenant user who do not have admin role is not permitted to call keystone's endpoints API. What user do you use in your test?

I used admin. However, all users who can login can get their own endpoints(catalog). 
There should be no limitation to user.

zhaoqin-github:
> "not permitted" means what? login failed? please check the tenant id is right. Also need to check if the user can access kinds of resources(nova neutron ...), that means RBAC is setted.

For example, in Queens env, I create a tenant named 'newman' which include a member named 'newman'. At first, in Horizon UI, I only add '_member_' role to that user. That user can call Nova, Neutron, Glance API without any problem. But when that user calls Keystone services and endpoints API, it get 403 error. After I add 'admin' role to that user, it can call services and endpoints API.

zongzw:
> > "not permitted" means what? login failed? please check the tenant id is right. Also need to check if the user can access kinds of resources(nova neutron ...), that means RBAC is setted.
> 
> For example, in Queens env, I create a tenant named 'newman' which include a member named 'newman'. At first, in Horizon UI, I only add '_member_' role to that user. That user can call Nova, Neutron, Glance API without any problem. But when that user calls Keystone services and endpoints API, it get 403 error. After I add 'admin' role to that user, it can call services and endpoints API.

Got it and it's true..
I tested with my `testTenant` and `testuser` the role is `_Member_` on both our Queens and my self platform xx.yyy.11.184. 
Seems our queens has some problem?

zhaoqin-github:
> Got it and it's true..
> I tested with my `testTenant` and `testuser` the role is `_Member_` on both our Queens and my self platform xx.yyy.11.184.
> Seems our queens has some problem?

Maybe that is not an API which is open to generic user.. Let me change my test code.

PR #194, opened 2019-06-19
Add checkAndWait just after installas3 and deploy partition.

zongzw:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -887,16 +879,9 @@ export class AdcStateCtrlr {
...
-      if (code !== partition) {
-        return false;
-      }
-    } catch (error) {
-      return false;
-    }
-    return true;
+    let resObj = await bigipMgr.getPartition(partition);
+    let code = JSON.parse(resObj)['body'][0]['name'];
+    return code === partition;
```
jikui: do we need to try to catch up the exception?  the getPartition function might throw some exceptions?

zongzw: To the functions `partitioned` `installed` ...(these check functions), we only expect 2 results: `true` and `non-true` for now. 
We don't care `non-true` is either false or exception because the outside caller can help us to handle the exception. 
So we need not catch the exception here to make the process simpler.

jikui: ok. get it. I am ok if we definitely don't care about the "non-true" case.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -253,14 +235,23 @@ export class AdcController extends BaseController {
...
+            status: AdcState.PARTITIONERR,
+            lastErr: `${AdcState.PARTITIONERR}: Fail to create partition`,
+          }),
+      );
     } catch (err) {
       this.logger.error(`Creating partition ${tenantName} Error.`);
       await this.serialize(adc, {
         status: AdcState.PARTITIONERR,
-        lastErr: `${AdcState.PARTITIONERR}: Fail to create partition`,
+        lastErr: `${AdcState.PARTITIONERR}: Fail to create partition: ${err}`,
```
jikui: Will the err be converted to string automatically?

zongzw: Changed it. thx.

jikui: Went through the changes and thanks for adding the partition related UT. :-)

PR #207, opened 2019-06-27
Fix adc.management initiation bug.

zongzw:


zongzw:
@haoshen61 After manual test on OS, at L584, the original code does not behave as expected. Need your review.

PR #212, opened 2019-06-28
use the path in bigip for the wafpolicy reference

jikui:


app/waf/src/controllers/wafpolicy.controller.ts
```diff
@@ -182,7 +182,7 @@ export class WafpolicyController extends BaseController {
         wafpolicy.id,
       );
     } catch (error) {
-      throw new HttpErrors.unprocessableEntity(
+      throw new HttpErrors.UnprocessableEntity(
         'upload wafpolicy to asg service failed',
```
Niklaus-xie: what is this line changed for?

Niklaus-xie: ok. there is the upper case there.

PR #213, opened 2019-06-28
add port security enabled

Niklaus-xie:


zongzw:
Not clear why do we need setting this.

Niklaus-xie:
4 connectivity

PR #220, opened 2019-07-05
Refactor the create / setup api.

zongzw:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -341,11 +355,27 @@ export class AdcController extends BaseController {
       tenantId: await this.tenantId,
     });
 
-    if (adc.type === 'HW' && !(await this.untrustAdc(adc))) {
-      throw new HttpErrors.UnprocessableEntity('Fail to untrust device');
+    if (adc.type === 'HW') {
```
zhaoqin-github: I think we can attempt to untrust HW or VE before delete it, right? If no trustedDeviceId, it can exit gracefully. And next setup can be deleting VE instance and ports from OpenStack. The last step is delete HW or VE db record. Correct?

app/waf/src/controllers/adc.controller.ts
```diff
@@ -441,25 +471,8 @@ export class AdcController extends BaseController {
...
-          this.deleteOn(adc, addonReq);
-          return {id: adc.id};
-        } else
-          throw new HttpErrors.UnprocessableEntity(
-            `Not ready for bigip VE to : ${AdcState.RECLAIMED}`,
-          );
-
       case 'setup':
-        if (await this.adcStCtr.readyTo(AdcState.DOINSTALLED)) {
+        if (await this.adcStCtr.readyTo(AdcState.TRUSTED)) {
```
zhaoqin-github: Can we also change resource action API to /adcs/{adcId}/setup at this time?

zongzw: This PR is used to  reorganize the API implementation. 
I prefer to make it done in other PRs by someone.

zhaoqin-github: Agree.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -342,11 +356,27 @@ export class AdcController extends BaseController {
       tenantId: await this.tenantId,
     });
 
-    if (adc.type === 'HW' && !(await this.untrustAdc(adc))) {
-      throw new HttpErrors.UnprocessableEntity('Fail to untrust device');
+    if (adc.type === 'HW') {
```
jikui: For the hw branch, we may need to modify further? If yes, could we just skip the change this time?

app/waf/src/services/bigip.service.ts
```diff
@@ -258,16 +253,13 @@ export class BigIpManager {
   }
 
   async installDO(): Promise<string> {
-    // create the body with following format.
-    // await this.mustBeReachable();
+    await this.mustBeReachable();
```
jikui: we have checked if the bigip is reachable or not in the uploadDO. Do we still need to check this in installDO?
If you are intended for the consistence about the coding style, I am also ok with this.

app/waf/src/controllers/adc.controller.ts
zongzw: HW branch is kept as it is now except adding the `await this.adcRepository.deleteById(id);` to keep it in the same behavior with VE. 
(You may notice the combined conditions divided into 2 here, however, they are actually in the same logic.)

app/waf/src/services/bigip.service.ts
```diff
@@ -292,7 +283,7 @@ export class BigIpManager {
       let checkFunc = async () => {
         return await impFunc().then(resObj => {
           let status = resObj['status'];
-          if (status !== 'FAILED') return true;
+          if (status === 'FINISHED') return true;
```
jikui: This condition is better than the original one.
Thanks

app/waf/src/services/bigip.service.ts
zongzw: Thanks. Let's keep it as it is. Still in finding the unexpected mustReachable issue..

zongzw:
Tested again, create and delete work OK for now.

PR #221, opened 2019-07-05
DO 1.5

Niklaus-xie:


Niklaus-xie:
yes, I know we might have DO container removed, but I don't want to depend on other PRs. If those other PRs don't get merged soon, we'll have dependencies and both user stories can not close. So I am still submitting this PR here, and we can remove unnecessary codes once those other PRs are merged.

zongzw:
Got it. Let's make it done in another PR to migrate DO functions from container to BIG-IP.

PR #227, opened 2019-07-08
install DO

jikui:
Install DO rpm into bigip


jikui:
Please help to review the diff and I am working on correcting the two UT cases now.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -751,6 +766,56 @@ export class AdcController extends BaseController {
...
+      ipAddr: cnct.ipAddress,
+      port: cnct.tcpPort,
+    });
+
+    try {
+      let resObj = await bigipMgr.getDOStatus();
+      let code = JSON.parse(resObj)['body'][0][0]['result']['status'];
+      return code === 'OK';
+    } catch {
+      console.log('DO URI has not been registered.');
```
jikui: DO URI has been registered.!

app/waf/src/controllers/adc.controller.ts
```diff
@@ -751,6 +753,56 @@ export class AdcController extends BaseController {
...
+      // check if do is already installed.
+      let cnct = adc.management.connection!;
+      let bigipMgr = await BigIpManager.instanlize({
+        username: cnct.username,
+        password: cnct.password,
+        ipAddr: cnct.ipAddress,
+        port: cnct.tcpPort,
+      });
+
+      if ((await this.isDOReady(adc)) === false) {
```
zongzw: Can we ignore the check, proceed the upload-install directly?

app/waf/src/datasources/bigip.datasource.json
```diff
@@ -25,6 +25,53 @@
           "cred64en"
         ]
       }
+    },
+    {
```
zongzw: It's OK to do it like this. 
Another option is to combine the uploadDO and installDO together, because they are both POST. 
We can organize `headers` within our code(in the uploadDO and installDO functions.)

It's OK for now.

app/waf/src/services/bigip.service.ts
```diff
@@ -205,6 +213,98 @@ export class BigIpManager {
...
+      let response = await this.bigipService.uploadDO(
+        url,
+        this.cred64Encoded,
+        fstats.size - 1,
+        fstats.size,
+        buffer,
+      );
+      let resObj = JSON.stringify(response);
+      return resObj;
+    } catch {
```
zongzw: it's better to `catch(error) {throw new Error(Upload DO RPM file error.: ${error});}`, so that we can know better why. 

Another better option is not catch it, and make it throw to upper caller(let upper caller to handle it.)

app/waf/src/controllers/adc.controller.ts
jikui: I prefer to keep this logic this time and we may discuss this further.

app/waf/src/services/bigip.service.ts
jikui: Sure, will make the change accordingly.

Thanks

app/waf/src/controllers/adc.controller.ts
jikui: prefer to keep this logic this time and we could discuss this further.

app/waf/src/datasources/bigip.datasource.json
jikui: Let's keep this at this time and discuss this further later.

app/waf/src/services/bigip.service.ts
```diff
@@ -205,6 +213,102 @@ export class BigIpManager {
...
+
+      return await checkAndWait(checkFunc, 60).then(
+        async () => {
+          return await impFunc().then(resObj => {
+            let status = resObj['status'];
+            return status;
+          });
+        },
+        () => {
+          throw new Error('No configsync IP');
```
zongzw: 'No configsync IP' should change to error details.

app/waf/test/fixtures/datasources/testrest.datasource.ts
```diff
@@ -1018,6 +1018,88 @@ export const StubResponses = {
     };
   },
 
+  bigipDOInfo200: (state: string | 'OK') => {
```
zongzw: If you want to default state to be OK, it may be: `state: string = 'OK'`

scripts/start_all.sh
```diff
@@ -7,9 +7,15 @@ cdir=`cd $(dirname $0); pwd`
     ./download_asg_dependencies.sh
 )
 
+(
+    cd $cdir/../dependiences
+    ./download_do.sh
```
zongzw: download_do.sh is under `scripts`, not `dependencies`.

app/waf/test/fixtures/datasources/testrest.datasource.ts
jikui: accept it

> If you want to default state to be OK, it may be: `state: string = 'OK'`
yes. will make the change accordingly

app/waf/src/services/bigip.service.ts
jikui: > 'No configsync IP' should change to error details.

yes, will make the change accordingly

scripts/start_all.sh
jikui: made the change accordingly.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -482,6 +480,10 @@ export class AdcController extends BaseController {
   }
 
   private async setupOn(adc: Adc, addon: AddonReqValues): Promise<void> {
+    // install DO
+    if (await this.adcStCtr.readyTo(AdcState.DOINSTALLED))
```
Niklaus-xie: seems to me it does not need to check twice here.  see line #462 and #484

jikui: accept this and I will make the change accordingly

PR #228, opened 2019-07-08
WIP: an example

Niklaus-xie:
Other points to consider:
1 find other one-to-many and modify
2 exception handling: pass invalid uuid and handle 500 Internal server Error
3 test cases & test


Niklaus-xie:
has been added into qinglong's commit, this can be dropped.

PR #231, opened 2019-07-09
Add dependency files to repo.

zongzw:


zongzw:
> Why do we need to save those plugin binaries to our code repo? Why not download them from their download link?

I'm not sure of this. We may need to release it as part of WAFAAS, which are the dependencies. 
We provide the download scripts to do that(scripts/download_do.sh) if they do not exist. 

We may not download them when starting WAFAAS because, we cannot control:
1. network connection. 
2. remote target's version correctness/validation.

zongzw:
> The part of changing the DO package name from a fixed one to an available is ok for me.

Tested this PR, works well. FYI.

PR #237, opened 2019-07-11
add ssh key param

Niklaus-xie:


app/waf/src/services/do.service.ts
```diff
@@ -325,6 +325,28 @@ export class OnboardingManager {
...
+          userType: 'root',
+          oldPassword: obData.management.connection!.rootPass,
+          newPassword: obData.management.connection!.rootPass,
+          keys: [obData.compute.sshKey!],
+        };
+
+        this.logger.debug('Add users part.');
+        return Object.assign(target, {root: userInfo});
+      } else {
+        this.logger.debug(`No provision found.`);
```
zongzw: Does it mean `No sshkey found.`?

Niklaus-xie: yes, I meant to log 'no provisioning of sshKey found'.

zongzw: why not changing it? 
From that statement 'No provision found.' in log, no one knows it refers to sshkey.

PR #240, opened 2019-07-11
fix one-to-many issue and test it

zhangqinglong235:
Original code returns a wrong list.

### DIFF
--- a/app/waf/src/controllers/rule.controller.ts
+++ b/app/waf/src/controllers/rule.controller.ts
@@ -20,6 +20,7 @@ import {
   Count,
   CountSchema,
   Where,
+  EntityNotFoundError,
 } from '@loopback/repository';
 import {
   post,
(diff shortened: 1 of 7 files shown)


app/waf/src/controllers/rule.controller.ts
```diff
@@ -299,11 +304,15 @@ export class RuleController extends BaseController {
     ruleId: string,
     @param(Schema.pathParameter('actionId', 'Action resource ID'))
     actionId: string,
-  ): Promise<CollectionResponse> {
+  ): Promise<Response> {
     const data = await this.ruleRepository
       .actions(ruleId)
       .find({where: {id: actionId}}, {tenantId: await this.tenantId});
-    return new CollectionResponse(Action, data);
+    if (data.length === 0) {
```
zongzw: Does the `findById` be better? so that we need no if-else.

Niklaus-xie: using findById also seems fine. the only thing is that findById seems to be only using one of the 2 ids from the URL. (although we can indeed use only one of the 2 ids to identify, because each id is already unique). For now, this way is also fine to me.

zongzw: It's up to you. I was just thinking that `findById` is cleaner. 
The 2 id concern may not the problem, I think. See the following: 

```
  @get(prefix + '/adcs/{adcId}', {
    responses: {
      '200': Schema.response(Adc, 'Successfully retrieve ADC resource'),
      '404': Schema.notFound('Can not find ADC resource'),
    },
  })
  async findById(
    @param(Schema.pathParameter('adcId', 'ADC resource ID')) id: string,
  ): Promise<Response> {
    let data = await this.adcRepository.findById(id, undefined, {
      tenantId: await this.tenantId,
    });
    return new Response(Adc, data);
  }
```

Am I missing some of your points?

zongzw: querying with 2 ids together should be required. That would help check the relation between rule and action.

PR #242, opened 2019-07-12
Add as3 rpm package and download script to repo.

zongzw:


zongzw:
during setup test, install as3 usually failed: timeout or download failed. (the download logic within TrustExtension is not stable but support file://). 
The as3 rpm package is big ~13M. 
So installing it from local should be controllable than from remote f5.com/github.

zongzw:
> I like this change, but do not like to archive rpm file to our code repo... Git may save several copy of rpm files in the future, so that git clone speed will become slow, and the CI test in lab machine will require much more time to clone code repository...

OK, let me make a little change to remove all rpms: asg-extensions / do / as3.

PR #243, opened 2019-07-13
Fix err state cannot be reclaimed and add its unit test.

zongzw:


zongzw:
@zhaoqin-github need to review again(solved the code conflict.)
Tested just now. functions work OK.

PR #246, opened 2019-07-15
error terminate using checkAndWait for isTrusted

haoshen61:

PR #248, opened 2019-07-16
distinguish between trust timeout and error

haoshen61:


haoshen61:
> It should be a little better. Is it possible to save ASG API error message in lastError property?

Not at this moment. Neither the original checkAndWait nor the current one is designed to carry error messages. This PR focuses on refining #246 and we could work on your idea later.

PR #249, opened 2019-07-17
enable https for adcaas

zhang-shengping:


app/waf/certs/domain.crt
```diff
@@ -0,0 +1,21 @@
+-----BEGIN CERTIFICATE-----
```
zongzw: 1. We may need to create a script names \<repo home>/scripts/create_server_keypair.sh, to create it dynamically if they do not exist. 
2. The folder can be \<repo home>/data/certs/domain.[crt|key]. 
3. Also, you may need to change \<repo home>/app/waf/Dockerfile and \<repo home>/deploy/docker-compose.yml file to adept your change(mount the folder \<repo home>/data/certs/ into container). 
4. 443 is system protected port, non-privileged user may not run process to listen on it, we may need to change it > 1024. 
5. The environments `ENABLE_HTTPS` `CERT_KEY`.. should be added to \<repo home>/deploy/appcluster.rc, also should be prefixed with 'WAF_'.

app/waf/index.js
```diff
@@ -14,22 +14,36 @@
...
       openApiSpec: {
         // useful when used with OASGraph to locate your application
         setServersFromRequest: true,
       },
     },
   };
+
+  if (process.env.ENABLE_HTTPS) {
+    config.rest['protocol'] = 'https';
+    config.rest['port'] = +process.env.WAF_APP_HTTPS_PORT || 443;
```
zongzw: Why not reuse the `WAF_APP_HTTPS_PORT`?

zhang-shengping: for differentiating to (HTTP port configuration) WAF_APP_PORT

zongzw: I think it have no relation with protocol setting. 
User can start either https or http service over `WAF_APP_PORT`. 
When you add one more `WAF_APP_HTTPS_PORT`, user may be confused about our program's behavior when both `WAF_APP_HTTPS_PORT` and `WAF_APP_PORT` are set.

app/waf/index.js
```diff
@@ -14,22 +14,36 @@
...
+    config.rest['key'] = fs.readFileSync(
+      process.env.CERT_KEY || './certs/domain.key',
```
zhaoqin-github: We are shipping this app in container image. A container image may not ship cert files. Usually user need to provide their own cert files. So you may need a volume path for them to inject cert files.

zhang-shengping: Yes, I agree.

data/certs/README.txt
```diff
@@ -0,0 +1,4 @@
+The certificate domain.cert and private key domain.key for test only!
```
zongzw: Could you add the `.gitignore` file(or add items to it if exists) to remove this crt/key file checkins. 
We should better add a script under scripts folder to generate the cert/key file dynamically instead of putting the 2 files there and saying "for test only".

app/waf/index.js
```diff
@@ -14,22 +14,44 @@
...
       openApiSpec: {
         // useful when used with OASGraph to locate your application
         setServersFromRequest: true,
       },
     },
   };
+
+  if (process.env.WAF_ENABLE_HTTPS === 'true') {
+    config.rest['protocol'] = 'https';
+    config.rest['port'] = +process.env.WAF_APP_PORT || 443;
```
zongzw: I think we needn't change the port, keep it as it is, using 3000 is good.

deploy/appcluster.rc
```diff
@@ -33,3 +33,9 @@ AS3_RPM_URL=file:///var/dependencies/f5-appsvcs-3.10.0-5.noarch.rpm
 
 ASG_HOST=asg-server
 ASG_PORT=443
+
+WAF_CERT_KEY= /var/data/certs/domain.key
+WAF_CERT_CRT= /var/data/certs/domain.crt
+WAF_ENABLE_HTTPS=false
+# WAF_ENABLE_HTTPS=
```
zongzw: Remove this line.

deploy/appcluster.rc
```diff
@@ -33,3 +33,9 @@ AS3_RPM_URL=file:///var/dependencies/f5-appsvcs-3.10.0-5.noarch.rpm
 
 ASG_HOST=asg-server
 ASG_PORT=443
+
+WAF_CERT_KEY= /var/data/certs/domain.key
```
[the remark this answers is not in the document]
zhang-shengping: I will check your reviews, a lot of lbaas agent issue comes today.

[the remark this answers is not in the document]
zongzw: OK, take your time for this PR.

data/certs/README.txt
zhang-shengping: I can remove the example certificates and key, users will know to create cert and key under this directory. However, there is no need to spend time on scripting and maintaining a script to create cert and key. No better script then Linux `openssl` command.

scripts/start_all.sh
```diff
@@ -11,7 +11,12 @@ cdir=`cd $(dirname $0); pwd`
 
 (
     cd $cdir/../deploy
+
+    certsdir=$cdir/../data/certs
+    openssl req -newkey rsa:2048 -nodes -keyout $certsdir/domain.key -x509 -days 365 -out $certsdir/domain.crt -subj "/C=CN/ST=BJ/L=BJ/O=Example/OU=IT/CN=f5-adcaas.com/emailAddress=f5-adcaas@example.com"
```
zongzw: Generate the key pair every time when `start_all.sh`?

zongzw: Could you please change `@example.com` `Example`? we are not an example. 
`f5-adcaas.com`??

zongzw: How about `openssl req -newkey rsa:2048 -nodes -keyout $certsdir/domain.key -x509 -days 365 -out $certsdir/domain.crt -subj "/C=CN/ST=BJ/L=BJ/O=F5Networks"`. 
We only use it as traffic encryption.

zhang-shengping: I think it is an example, users will use there own cert and key in real production env.

zongzw: Generate the key pair every time when start_all.sh?

zhang-shengping: Yes，I really do not what to keep a key file, it is not secure either for test or example env, everyone can download the private key right?

zongzw: No one can download key pairs, I think. 

The problem of generating kp everytime is you may overwrite user's ones if they don't change the path. 
Why not make it re-doable. Like download_*. They can be called any times without any negative effect.

PR #251, opened 2019-07-17
Make password generation configurable.

zongzw:


zongzw:
Test result: 
```
            "connection": {
                "username": "admin",
                "password": "k3jApA?/|n4A",
                "rootPass": "-FOc^OT2QFO?",
                "tcpPort": 443,
                "ipAddress": "10.250.15.158"
            },
```

zongzw:
@olvandeng, will we re-define the acceptance criteria again. 
@zhaoqin-github We are reaching to the end of the sprint. Why not you post your suggestion earlier?

zongzw:
> seems to me this does not work. did you test the setup process,etc? need to take care of some hardcoded passwords so that it does not break anything.

Of course, I tested it and it works well. Just now I tested it again, the processes till to installing as3 are OK.   (partition creation fails, but it seems no matter with password, because we have created the trust. )

What's the issue posted from it?

Niklaus-xie:
> > seems to me this does not work. did you test the setup process,etc? need to take care of some hardcoded passwords so that it does not break anything.
> 
> Of course, I tested it and it works well. Just now I tested it again, the processes till to installing as3 are OK. (partition creation fails, but it seems no matter with password, because we have created the trust. )
> 
> What's the issue posted from it?

that is exactly the problem. it can create the partition correctly before this patch,  but it can not do that now with this patch.  so you should find out why the password generation causes that issue, otherwise, this patch can not be merged because it introduces issues(the setup api does not succeed).  As i mentioned, it is probably because of some hardcoded passwords earlier.

zongzw:
> > > seems to me this does not work. did you test the setup process,etc? need to take care of some hardcoded passwords so that it does not break anything.
> > 
> > 
> > Of course, I tested it and it works well. Just now I tested it again, the processes till to installing as3 are OK. (partition creation fails, but it seems no matter with password, because we have created the trust. )
> > What's the issue posted from it?
> 
> that is exactly the problem. it can create the partition correctly before this patch, but it can not do that now with this patch. so you should find out why the password generation causes that issue, otherwise, this patch can not be merged because it introduces issues(the setup api does not succeed). As i mentioned, it is probably because of some hardcoded passwords earlier.

Thanks for finding this. It's indeed the problem. 
@jikui help fixed in PR https://github.com/F5Networks/f5-adcaas-openstack/pull/265. After rebasing the code, tested again. works ok for now.

Niklaus-xie:
> > > > seems to me this does not work. did you test the setup process,etc? need to take care of some hardcoded passwords so that it does not break anything.
> > > 
> > > 
> > > Of course, I tested it and it works well. Just now I tested it again, the processes till to installing as3 are OK. (partition creation fails, but it seems no matter with password, because we have created the trust. )
> > > What's the issue posted from it?
> > 
> > 
> > that is exactly the problem. it can create the partition correctly before this patch, but it can not do that now with this patch. so you should find out why the password generation causes that issue, otherwise, this patch can not be merged because it introduces issues(the setup api does not succeed). As i mentioned, it is probably because of some hardcoded passwords earlier.
> 
> Thanks for finding this. It's indeed the problem.
> @jikui help fixed in PR #265. After rebasing the code, tested again. works ok for now.
cool.

PR #255, opened 2019-07-22
Setup action API into URL Path

zhangqinglong235:
API URL and body change

### DIFF
--- a/app/waf/src/controllers/adc.controller.ts
+++ b/app/waf/src/controllers/adc.controller.ts
@@ -37,9 +37,7 @@ import {
 import {
   Adc,
   Tenant,
-  ActionsBody,
   ActionsResponse,
-  ActionsRequest,
   AS3PartitionRequest,
   as3Name,
 } from '../models';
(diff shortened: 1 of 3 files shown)


app/waf/src/models/adc.model.ts
```diff
@@ -191,14 +177,3 @@ export class ActionsResponse extends Entity {
   id: string;
 }
```
Niklaus-xie: need to delete  empty line at the end of this file. otherwise it won't pass tests.  after running "npm run prettier:fix", you should add those modifications by 'git add XXXX, git commit --amend, etc'.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -451,15 +449,13 @@ export class AdcController extends BaseController {
   //     "id": "11111111-2222-3333-4444-555555555555"
   //   }
   // }
-  @post(prefix + '/adcs/{adcId}/action', {
+  @post(prefix + '/adcs/{adcId}/setup', {
     responses: {
       '200': Schema.response(ActionsResponse, 'Adc Id for the actions.'),
```
zhaoqin-github: Not quite sure if we will still need this response body. Usually 204 with empty body will be OK for an asyncronous API. We can decide that later.

PR #258, opened 2019-07-25
add requestId in logger names

haoshen61:
JIRA 1546

### DIFF
--- a/app/waf/src/controllers/adc.controller.ts
+++ b/app/waf/src/controllers/adc.controller.ts
@@ -65,6 +65,7 @@ const prefix = '/adcaas/v1';
 export class AdcController extends BaseController {
   asgMgr: ASGManager;
   private adcStCtr: AdcStateCtrlr;
+  private logger = factory.getLogger('Unknown: controllers.adc');
 
   constructor(
     @repository(AdcRepository)
@@ -78,10 +79,12 @@ export class AdcController extends BaseController {
     protected reqCxt: RequestContext,
     @inject(CoreBindings.APPLICATION_INSTANCE)
     private wafapp: WafApplication,
-    private logger = factory.getLogger('controllers.adc'),
   ) {
     super(reqCxt);
-    this.asgMgr = new ASGManager(this.asgService);
+    if (reqCxt) {
+      this.asgMgr = new ASGManager(this.asgService, this.reqCxt.name);
+      this.logger = factory.getLogger(this.reqCxt.name + ': controllers.adc');
+    }
   }
 
   @post(prefix + '/adcs', {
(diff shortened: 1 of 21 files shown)


app/waf/src/controllers/adc.controller.ts
```diff
@@ -67,6 +67,7 @@ const prefix = '/adcaas/v1';
 export class AdcController extends BaseController {
   asgMgr: ASGManager;
   private adcStCtr: AdcStateCtrlr;
+  private logger = factory.getLogger('Unknown: controllers.adc');
```
zongzw: `private logger = factory.getLogger('Unknown: controllers.adc');` could be removed in my opinion, because we have the initialiation in the `constructor`. 
I suggest you remove L80 `{optional: true}`, because `RestBindings.Http.CONTEXT` must have been bound. 
Thus `if (reqCxt) {` is needless.

app/waf/src/controllers/adc.controller.ts
```diff
@@ -113,7 +116,7 @@ export class AdcController extends BaseController {
       userToken: await this.reqCxt.get(WafBindingKeys.Request.KeyUserToken),
       tenantId: await this.reqCxt.get(WafBindingKeys.Request.KeyTenantId),
     };
-    this.adcStCtr = new AdcStateCtrlr(adc, addonReq);
+    this.adcStCtr = new AdcStateCtrlr(adc, addonReq, this.reqCxt.name);
```
zongzw: Is it possible to combine `this.reqCxt.name` into `addonReq`? That means adding one more member to `AddonReqValues`, whose type is `string`, or `RequestContext` directly.

app/waf/src/controllers/declaration.controller.ts
```diff
@@ -100,6 +102,11 @@ export class DeclarationController extends BaseController {
     private wafapp: WafApplication,
   ) {
     super(reqCxt);
+    if (reqCxt) {
```
zongzw: Same to this one, how about removing `{optional: true}` when injecting.

app/waf/src/sequence.ts
```diff
@@ -50,6 +50,7 @@ export class MySequence implements SequenceHandler {
   ) {}
 
   async handle(context: RequestContext) {
+    this.logger = factory.getLogger(context.name + ': api.call');
```
zongzw: Seems this line can be moved into `constructor`. And then remove the constructor's injection of `logger`.

app/waf/src/services/bigip.service.ts
```diff
@@ -87,7 +92,7 @@ export class BigIpManager {
     let impFunc = async () => {
       let response = await this.bigipService.getInfo(url, this.cred64Encoded);
       let resObj = JSON.parse(JSON.stringify(response))['body'][0];
-      this.logger.debug(`get ${url} resposes: ${JSON.stringify(resObj)}`);
+      this.logger.debug(`get ${url} responses: ${JSON.stringify(resObj)}`);
```
zongzw: thanks.

PR #261, opened 2019-07-26
Spike: add EFK for adcaas (this code will not merge to master, unless future it is required)

zhang-shengping:
add EFK deploy for ADCAAS.

### DIFF
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -79,15 +79,41 @@ services:
         fluentd-address: "0.0.0.0:24224"
         tag: f5-do
 
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.2.0
+    container_name: F5-ADCAAS-ELASTICSEARCH
+    ports:
+      - 9200:9200
+    environment:
+      - discovery.type=single-node
+
   FLUENTD:
-    image: fluent/fluentd:latest
+    build: ./docker/fluent
+    image: f5-adcaas-fluent
     container_name: FLUENTD
     ports:
       - 24224:24224
+      - 24224:24224/udp
+      - 20001:20001/udp
     volumes:
       - ./fluentd.conf:/fluentd/etc/test.conf
+    depends_on:
+      - elasticsearch
+      - kibana
     environment:
       - FLUENTD_CONF=test.conf
+
+  kibana:
+    image: docker.elastic.co/kibana/kibana:7.2.0
+    container_name: F5-ADCAAS-KIBANA
+    depends_on:
+      - elasticsearch
+    ports:
+      - 5601:5601
+    expose:
+      - 5601
+    volumes:
+      - ./kibana.yml:/usr/share/kibana/config/kibana.yml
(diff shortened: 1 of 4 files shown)


deploy/docker-compose.yml
```diff
@@ -78,15 +78,42 @@ services:
         fluentd-address: "0.0.0.0:24224"
         tag: f5-do
 
+  elasticsearch:
```
zongzw: We may need to mount the data directory of elasticsearch out as a volume.

deploy/docker-compose.yml
```diff
@@ -78,15 +78,42 @@ services:
...
+  kibana:
+    image: docker.elastic.co/kibana/kibana:7.2.0
+    container_name: F5-ADCAAS-KIBANA
+    depends_on:
+      - elasticsearch
+    ports:
+      - 5601:5601
+    expose:
+      - 5601
+    volumes:
```
zongzw: We may need to mount kibana data directory as a volume so that user-created index data can be persisted.

deploy/docker-compose.yml
zhang-shengping: I do not know how to reuse the data from node elasticsearch to another elasticaserach or the data volume is usable or not, you probably can have a try and investigate more after this patch. 
I think the better solution is to set up a cluster, we can have a investigate after this sprint demo.
For more secure, we can put elasticserach on a server node (physical or virtual one).

deploy/docker-compose.yml
zhang-shengping: Yes, we can do that, we can have a investigate in next sprint about HA of Kibana, Elasticserach, Fluentd.

deploy/docker-compose.yml
```diff
@@ -78,15 +78,42 @@ services:
         fluentd-address: "0.0.0.0:24224"
         tag: f5-do
 
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.2.0
+    container_name: F5-ADCAAS-ELASTICSEARCH
+    ports:
+      - 9200:9200
+    expose:
```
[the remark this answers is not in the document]
zhang-shengping: Yes, we can. I will delete this.

deploy/fluentd.conf
```diff
@@ -1,16 +1,24 @@
...
-  @type stdout
+<match *.**>
+  @type copy
+  <store>
+    @type elasticsearch
+    host elasticsearch
+    port 9200
+    logstash_format true
+    logstash_prefix fluentd
+    logstash_dateformat %Y%m%d
```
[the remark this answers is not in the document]
zhang-shengping: yes, I checked, the default dateformat works well too for logging.

deploy/docker-compose.yml
```diff
@@ -78,15 +78,42 @@ services:
...
+      - 9200
+    environment:
+      - discovery.type=single-node
```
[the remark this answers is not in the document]
zhang-shengping: I not quite sure too.

PR #266, opened 2019-07-31
create certs directory for keep key and certs

zhang-shengping:


scripts/start_all.sh
```diff
@@ -13,6 +13,7 @@ cdir=`cd $(dirname $0); pwd`
     cd $cdir/../deploy
 
     certsdir=$cdir/../data/certs
+    mkdir $certsdir
```
zongzw: In fact, we have `source appcluster.rc` where `WAF_CERT_CRT` and `WAF_CERT_KEY` are defined. 
I think you can directly use these variables instead of refer/mkdir $certdir. 
Even not, replace `mkdir` with `mkdir -p`.

zhang-shengping: I think the variables in appcluster.rc is the path in a container, not outside, right?

zongzw: oh, yeah.. my fault. 
you may need to change `mkdir` to `mkdir -p`, or the second time to run `mkdir` will fail, isn't it?

zhang-shengping: it will not fail, it will give a reminder that the directory exists: `mkdir: /Users/pzhang/git/f5-adcaas-openstack/scripts/../data/certs: File exists`

PR #275, opened 2019-08-12
Connect to Beijing lab vpn

zhaoqin-github:


scripts/vpn_connect.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+INTERVAL=10
+RETRY=6
+
+/usr/local/pulse/PulseClient_x86_64.sh -h $VPN_HOST -u $VPN_USER -p $VPN_PASS -r Users &
```
zongzw: This command may leak credentials: 

```
Checking for missing dependency packages for command line client ...
executing command : /usr/local/pulse/pulsesvc -h 183.84.2.165 -u xxxx -p yyyyy -r Users
```
Should redirect the output into > /dev/null

zhaoqin-github:
> Could we run the vpn in container. connect VPN just for collect running result, and all working stuff should be run remotely in lab.
> 
> The ubuntu image has own 64M, small enough to download from dockerhub.

What is its size after you install the vpn client?

scripts/vpn_connect.sh
zhaoqin-github: done

PR #284, opened 2019-08-15
adc validation

Niklaus-xie:


app/waf/src/models/adc.model.ts
```diff
@@ -17,6 +17,110 @@
...
+    type: 'string',
+    format: 'ipv4',
+  },
+  floatingIp: {
+    type: 'string',
+    format: 'ipv4',
+  },
+};
+
+const adcNetworksProps = {
```
zhaoqin-github: The 'networks' property is a dictionary, so it allows user to input any key. For instance:

```
{
  "networks" {
    "mgmtABC" : {},
    "ha": {},
    "internal123": {},
    "external456" {}
  }
}
```

We can not type the exact key name in API schema, and should use another way to validate this dictionary.
https://swagger.io/docs/specification/data-models/dictionaries/

zhaoqin-github:
Let us merge this change now, and I will create another change to fix the validation of networks property.

zhaoqin-github:
Create #293

PR #288, opened 2019-08-16
Rename WAFAAS to ADCAAS

zhaoqin-github:


zhaoqin-github:
> If you name it like this, how do you name later ltmaas dnsaas?
> do you want to implement them all in the same container?
> How to apply the micro-service concept to our delivery?

Not sure about dnsaas. Our current code is mainly for ltmaas and wafaas. We can think about the naming, when we need another app, or need to demerge this app. I think we can use a generic name in this stage to prepare for future renaming.

scripts/README.md
```diff
@@ -76,7 +76,7 @@ The following chart shows the container interactions:
                         +--------------+---------------+
                         |                              |
      +-------------------------------------+     +------------+
-     |       LTMAAS/DNSAAS/..(future)      |     |   WAFAAS   |
+     |       LTMAAS/DNSAAS/..(future)      |     |   ADCAAS   |
```
zongzw: We may need to keep this as WAFAAS.

zhaoqin-github: Done

PR #335, opened 2019-09-18
Use DO rpm instead of container

zhaoqin-github:


app/waf/src/controllers/adc.controller.ts
```diff
@@ -744,9 +747,15 @@ export class AdcController extends BaseController {
...
             if (adc.license) return true;
-            return await bigipMgr.getLicense().then(license => {
-              return license.registrationKey === 'none';
-            });
+            return await bigipMgr.getLicense().then(
+              license => {
+                return license.registrationKey === 'none';
+              },
+              err => {
+                //Supress all errors, since iControl is restarting
```
netcodergit: Do we have  the other logs to record this error situation?

app/waf/src/controllers/adc.controller.ts
```diff
@@ -744,9 +747,15 @@ export class AdcController extends BaseController {
...
           let noLicensed = async () => {
             if (adc.license) return true;
-            return await bigipMgr.getLicense().then(license => {
-              return license.registrationKey === 'none';
-            });
+            return await bigipMgr.getLicense().then(
+              license => {
+                return license.registrationKey === 'none';
+              },
+              err => {
```
jikui: Do we have the other places to record this kind of error status?

app/waf/src/models/adc.model.ts
```diff
@@ -220,6 +220,20 @@ export class Adc extends CommonEntity {
   constructor(data?: Partial<Adc>) {
     super(data);
   }
+
+  getBasicAuth(): string {
+    return Buffer.from(
+      `${this.management.connection!.username}:${
+        this.management.connection!.password
```
jikui: This username and password is used for storing the username/password for the openstack? If yes, when the timeout occurs, will these information be updated as well?

app/waf/src/models/adc.model.ts
```diff
@@ -220,6 +220,20 @@ export class Adc extends CommonEntity {
   constructor(data?: Partial<Adc>) {
     super(data);
   }
+
```
zongzw: It's not a good way to make a model to have behaviors, that may be a bad implementation design. 
Loopback4 is a MVC framework. Model is the place to map database layout. 

The similar situation happens to application.model.ts and common.model.ts. getAS3Declaration staff.

app/waf/src/services/do.service.ts
```diff
@@ -512,6 +514,19 @@ export class OnboardingManager {
           return resObj[0]['result']['code'] === 200;
         },
         reason => {
+          // Wait for iControl LX restarting
```
zongzw: What's other possible statusCode?

app/waf/test/fixtures/controllers/mocks/mock.bigip.controller.ts
```diff
@@ -66,6 +66,18 @@ export class MockBigipController extends MockBaseController {
     );
   }
 
+  @post('/mgmt/shared/declarative-onboarding')
+  async doOnboard(): Promise<object> {
+    return await ResponseWith.do_post_mgmt_shared_declaration_onboarding!();
```
zongzw: rename it to bigip_

app/waf/test/fixtures/controllers/mocks/mock.bigip.controller.ts
```diff
@@ -66,6 +66,18 @@ export class MockBigipController extends MockBaseController {
...
+  }
+
+  @get('/mgmt/shared/declarative-onboarding/task/{taskId}')
+  async doTaskResult(
+    @param.path.string('taskId') taskId: string,
+  ): Promise<object> {
+    return await ResponseWith.do_get_mgmt_shared_declaration_onboarding_task_taskId!();
```
zongzw: rename it to bigip_

app/waf/test/unit/service.OnboardingManager.ts
```diff
@@ -93,7 +93,8 @@ describe('test OnboardingManager', async () => {
   });
 
   it('onboard: ok', async () => {
-    let adc = <Adc>createAdcObject();
+    let adc = new Adc(createAdcObject());
+    doMgr.config.endpoint = adc.getDoEndpoint();
```
zongzw: This should be passed in when doMgr is instantiated.

app/waf/src/models/adc.model.ts
zhaoqin-github: They are username and password of bigip. Needn't to update them.

app/waf/src/services/do.service.ts
zongzw: That means there's a big possibility to change this code part again when encountering new status need-to-wait in the future. maybe again and again?

Could we ignore details status, wait until timeout or success? Because till now we don't know what kinds of statuses need us to quit. 

Seems restnoded has the ability to recover(restart self), there may be little possibility of quit because of server error(5XX) or do install failed(4xx).

Maybe it's more suitable to write it like this: 

```
if (some quit reason) {
    return Promise.reject('reason')
}
if(other quit) {
    // quit immediately. 
}

// default -> wait..
```

app/waf/test/fixtures/controllers/mocks/mock.bigip.controller.ts
zhaoqin-github: done

app/waf/test/unit/service.OnboardingManager.ts
zhaoqin-github: done

app/waf/test/fixtures/controllers/mocks/mock.bigip.controller.ts
zhaoqin-github: done

app/waf/src/services/do.service.ts
zhaoqin-github: Your first two 'if' action are equivalent.
I will propose
```
if (some reason) {
  // wait
}
// default quit immediately.
```
as we used to do, because the code can not enumerate unexpected errors. Under most circumstance, the program need to be broken by the unexpected errors, unless we specify to tolerate them.

app/waf/src/models/adc.model.ts
zhaoqin-github: Those two are property made by some calculation. This is a very common case. An object's can have some properties originate from the calculation with some other properties. I do not think moving those code can benefit the program.

zongzw: I don't think so. However, it's up to yo.

zhaoqin-github:
@zongzw made a very small change at line https://github.com/F5Networks/f5-adcaas-openstack/blob/8835f822c349b5cf5a1a719300e65d6216f8acbe/app/waf/src/services/do.service.ts#L474

Please review again.

PR #338, opened 2019-09-23
Do not delete ADC from db when reclaiming error occurs

zhaoqin-github:


zhaoqin-github:
> Personally speaking, asynchronizingly accessing DB to save status should be better.

This change is willing to fix the bug that the db record is deleted before you update its content. And I notice a bigger bug today...  #340

PR #342, opened 2019-09-27
TMA - remove credential information

zhaoqin-github:


zongzw:
The real big deal should be postman test environments part. 
How will we deal with it? @zhaoqin-github

PR #343, opened 2019-09-27
Catch ASG error of failing to upload WAF policy

zhaoqin-github:


app/waf/src/controllers/wafpolicy.controller.ts
```diff
@@ -227,9 +227,11 @@ export class WafpolicyController extends BaseController {
...
       );
     } catch (error) {
-      throw new HttpErrors.unprocessableEntity(
-        'check wafpolicy from asg service failed',
-      );
+      if (error.code === 404) {
+        throw new HttpErrors.NotFound(`ASG error: ${error.message}`);
+      } else {
+        throw new HttpErrors.UnprocessableEntity(`ASG error: ${error.message}`);
+      }
```
zongzw: `ASG error` may be not accurate. How do we know if fails because of ASG not because of some reason else.
Furthermore, I think this change's benefit is adding `${error.message}` in the error response. `if (error.code === 404)` is useless.

PR #352, opened 2019-10-15
Add LTMaaS iRule configuration.

zongzw:


app/waf/src/controllers/declaration.controller.ts
```diff
@@ -172,6 +175,29 @@ export class DeclarationController extends BaseController {
         defs: defs,
       };
     }
+
+    if (service.iRules && service.iRules.length > 0) {
+      let refs: (string | object)[] = [];
+      let defs: {[k: string]: object} = {};
+
+      for (let i of service.iRules) {
```
zhaoqin-github: Maybe we can add a 'findByIds()' in repository class to load all iRules in one db call. Let's do it later.

zongzw: I need to check the iRule ID one by one so that to determine using 'bigip' or 'use'. 
Can `findByIds` work when iRule with not-exist-ID?

app/waf/src/models/irule.model.ts
```diff
@@ -0,0 +1,91 @@
...
+      },
+    },
+    as3: {},
+  })
+  label?: string;
+
+  @property({
+    type: 'string',
+    required: false,
+    default: '',
```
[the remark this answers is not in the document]
zongzw: agree. I prefer to use `default ''`, because it can be returned as a response property.

app/waf/src/models/service.model.ts
```diff
@@ -108,6 +108,27 @@ export class Service extends CommonEntity {
...
+        items: {
+          type: 'string',
+        },
+      },
+    },
+    as3: {
+      type: 'extends',
+    },
+  })
+  iRules?: string[];
```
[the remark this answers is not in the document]
zongzw: Done

app/waf/src/controllers/declaration.controller.ts
zhaoqin-github: I think it should return a set of result, which includes the iRules in db. Then we can go through the ids, if it is not in the result set, it can use 'bigip' to load iRule from '''/Common''' partition, right? Let's think that later.

PR #354, opened 2019-10-16
Unreachable license

zhaoqin-github:


app/waf/test/fixtures/datasources/testrest.datasource.ts
```diff
@@ -24,6 +24,7 @@ export enum RestApplicationPort {
   IdentityUser = 5000,
   Onboarding = 8081,
   ASG = 8443,
+  BigIq = 8888,
```
zongzw: // in order, please.

app/waf/test/fixtures/datasources/testrest.datasource.ts
```diff
@@ -2186,6 +2187,33 @@ Reason:
...
+    return {
+      status: 'FAILED',
+    };
+  },
+
+  bigiqRevokeTaskFinished200: () => {
+    return {
+      status: 'FINISHED',
+    };
+  },
```
zongzw: Please post the real responses here.

app/waf/src/services/bigiq.service.ts
```diff
@@ -0,0 +1,167 @@
...
+import {BigIqDataSource} from '../datasources';
+import {factory} from '../log4ts';
+import {Logger} from 'typescript-logging';
+import {checkAndWait} from '../utils';
+
+const BIGIQ_HOST: string = process.env.BIGIQ_HOST || 'localhost';
+const BIGIQ_PORT: number = Number(process.env.BIGIQ_PORT) || 8888;
+const BIGIQ_USERNAME: string = process.env.BIGIQ_USERNAME || 'admin';
+const BIGIQ_PASSWORD: string = process.env.BIGIQ_PASSWORD || 'admin';
+const BIGIQ_POOL: string = process.env.BIGIQ_POOL || 'unknown';
```
zongzw: Throw exception here instead of giving a default value. 
With the given default value, ADCaaS may run into uncontrollable state.

app/waf/src/services/bigiq.service.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  value(): Promise<BigIqService> {
+    return getService(this.dataSource);
+  }
+}
+
+export class BigIqManager {
+  private service: BigIqService;
+  private logger: Logger;
+  private static token: BigIqToken;
```
zongzw: If you think BIG-IQ is a singleton, why not making `BigIqManager` static? 
By the way, I think static here is not a good idea.

app/waf/src/services/bigiq.service.ts
```diff
@@ -0,0 +1,167 @@
...
+  async revoke(body: object): Promise<void> {
+    let token = await this.login();
+    let revokeResp = await this.service.revoke(
+      BIGIQ_HOST,
+      BIGIQ_PORT,
+      token.token,
+      body,
+    );
+
+    let taskFinished = async (taskId: string): Promise<boolean> => {
```
zongzw: The parameter `taskId: string` is useless here, since you can capture `revokeResp.id` within the inline function. 

By the way, function `checkAndWait`'s 3rd parameter is useless, and it may lead to potential issues.

app/waf/src/services/bigiq.service.ts
```diff
@@ -0,0 +1,167 @@
...
+    return getService(this.dataSource);
+  }
+}
+
+export class BigIqManager {
+  private service: BigIqService;
+  private logger: Logger;
+  private static token: BigIqToken;
+
+  constructor(svc: BigIqService, reqId = 'Unknown') {
```
zongzw: No need to pass in. See `LicenseManager`'s instantiation.

app/waf/src/services/license.service.ts
```diff
@@ -133,7 +137,16 @@ export class LicenseManager {
 
   // // BIG-IQ ✓<-->x BIG-IP
   // private async licViaBIQ() { }
-  // private async unLicViaBIQ() { }
+  private async unLicViaBIQ(adc: Adc): Promise<void> {
+    let address = adc.management.connection!.ipAddress;
+    let mac = '';
+    for (let net of Object.keys(adc.networks)) {
+      if (adc.networks[net].type === 'mgmt') {
+        mac = adc.management.networks[net].macAddr!;
```
zongzw: Be careful that `adc.networks` exists but `adc.management.networks` doesn't. 
the reclaimFuncs.license in `deleteOn` has no `try..catch..`. When VE is powered partially, is it possible to make reclaiming always failed because of "reading `adc.management.networks` undefined error"?

app/waf/src/services/license.service.ts
```diff
@@ -133,7 +137,16 @@ export class LicenseManager {
...
+      }
+    }
+    await this.biqMgr.revokeLicense(address, mac);
```
zongzw: Without await, `return this.biqMgr.revokeLicense(address, mac);` should be OK.

app/waf/src/services/bigiq.service.ts
zhaoqin-github: done

app/waf/src/services/license.service.ts
zhaoqin-github: done

app/waf/src/services/bigiq.service.ts
zhaoqin-github: If we do it in that way, we will need to create bigiq manager instance in controller layer, so that the the flow will have some other changes. Please confirm you really hope to make this change in this commit, or delay it to another commit.

app/waf/test/fixtures/datasources/testrest.datasource.ts
zhaoqin-github: What information do you hope to write here?

app/waf/src/services/bigiq.service.ts
zhaoqin-github: I was not thinking bigiq is a singleton. I just hope not to login bigip too many times. In fact I am not sure what harm will this static property cause. If you can enumerate some examples, we can consider a better way. Anyway, changing it to static will depends on it initialization implementation. Let decide them as a whole, delay or change it in this commit.

app/waf/test/fixtures/datasources/testrest.datasource.ts
zhaoqin-github: Other properties are useless. We needn't to add them.

app/waf/src/services/bigiq.service.ts
zhaoqin-github: The default values themselves is harmless. If we hope to ensure every required settings are there, we will need to have another piece of code to do this check when the program startup. However, we do not have that code now. Need to do that in another change set.

app/waf/src/services/license.service.ts
zhaoqin-github: If the VE is in power error state, we should not revoke license, right?

PR #358, opened 2019-10-18
Remove do container

zhaoqin-github:


app/waf/src/services/do.service.ts
```diff
@@ -118,15 +117,15 @@ export class OnboardingManager {
       );
 
     this.config = {
-      endpoint: process.env.DO_ENDPOINT!,
+      endpoint: '',
```
zongzw: I miss getting this point. Where does the code assign the `this.config.endpoint` before being used in ` async onboard(givenDoBody: TypeDOClassDO)`?

zhaoqin-github: You may look at https://github.com/F5Networks/f5-adcaas-openstack/pull/335

DO manager need to get ADC's DO endpoint url during init

zongzw: Should not a PR an integrated change?

PR #369, opened 2019-11-14
Assign license via BIQ-IQ

zhaoqin-github:


app/waf/src/models/adc.model.ts
```diff
@@ -217,16 +214,18 @@ export class Adc extends CommonEntity {
   })
   lastErr: ConfigTypes['lastErr'];
 
+  username: string = 'admin';
```
jikui: The username and password will be default?

app/waf/src/controllers/adc.controller.ts
```diff
@@ -153,6 +147,11 @@ export class AdcController extends BaseController {
           await runWithTimer(`${adc.id}.onboard`, () =>
             this.onboard(adc, addonReq),
           );
+
```
jikui: The logic is for the Hardware bigip?

app/waf/src/controllers/adc.controller.ts
```diff
@@ -116,7 +110,7 @@ export class AdcController extends BaseController {
       tenantId: await this.reqCxt.get(WafBindingKeys.Request.KeyTenantId),
       reqId: this.reqCxt.name,
     };
-    this.adcStCtr = new AdcStateCtrlr(adc, addonReq);
+    this.adcStCtr = new AdcStateCtrlr(adc, addonReq, this.asgMgr);
```
zongzw: It's needless to pass the 3rd parameter to `AdcStateCtrlr`. 
If you are concerning the unit tests and want to stub it up with Sinon, why not stub it with:
`let stubInst = sinon.stub(AdcStateCtrlr.prototype, "some-function"); .. stubInst.callfake(..); ..`.
Have you searched online for it, or it doesn't work for you?

app/waf/src/controllers/adc.controller.ts
```diff
@@ -153,6 +147,11 @@ export class AdcController extends BaseController {
           await runWithTimer(`${adc.id}.onboard`, () =>
             this.onboard(adc, addonReq),
           );
+
+        if (await this.adcStCtr.readyTo(AdcState.TRUSTED))
```
zongzw: Why adding `setup` to the poweron/onboarding process?

app/waf/src/models/adc.model.ts
zhaoqin-github: They are default value. They can be replaced by random value, if we specify ```VE_RANDOM_PASS=true```

app/waf/src/controllers/adc.controller.ts
zhaoqin-github: No logic for HW, because we postpone HW support.

PR #396, opened 2019-12-19
Import Token into release deploy

JingL1:


.travis.yml
```diff
@@ -60,6 +60,8 @@ jobs:
       deploy:
       # RELEASE ADCAAS 
       - provider: releases
+        api_key:
+          secure: "$GITHUB_TOKEN"
```
Niklaus-xie: maybe check the doc if this format is ok.

